PR #11, opened 2018-04-21
Optimize assets

luzfcb:
I usually use https://kraken.io/web-interface to optimize the image file size. Works well. Including for .gif files . ( free version limit the max file size to 1MB)


Irio:
Thanks @luzfcb

PR #20, opened 2018-04-25
Remove max argument from Campinas Scrapy contract

cuducos:
Probably they just published a new gazette and this might be breaking our CI


processing/data_collection/gazette/spiders/sp_campinas.py
```diff
@@ -33,7 +33,7 @@ def parse(self, response):
     def parse_month_page(self, response):
         """
         @url http://www.campinas.sp.gov.br/diario-oficial/index.php?mes=1&ano=2018
-        @returns items 22 22
+        @returns items 22
```
Irio: It seems that Campinas just published a new gazette in January... 🤔 

In this case, it's better to stick with the new number.

```
@returns items 23 23
```

cuducos: Done :tada:

PR #24, opened 2018-04-25
Add instructions for running the scrapper without docker

kerollaine:
I've found this to be quite useful, specially for beginners, but I don't know if there is a better way to debug without ditching Docker, so feel free to close without merging if that is the case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,3 +25,47 @@ Once the download and building processes are finished, the processing tasks shou
 ## Contributing
 
 If you are interested in fixing issues and contributing directly to the code base, please see the document [CONTRIBUTING.md](CONTRIBUTING.md).
+
+## Running the Project Without Docker
+
+Even if your environment is working with the setup described above - and it should be, because you'll probably want to test your work inside docker before sending PR's - sometimes you'd like to run the project without docker for some reason, for example to have an easier time debugging python code. In this section we will explain one way to do this. Note that you'll need the database container for this to work, i.e. be sure to follow through the normal docker setup before following the steps below.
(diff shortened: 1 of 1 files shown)


Irio:
I feel that maintaining multiple instructions at this time of the project might take too much effort, rather than having a single way of running everything, which is heavily tested by all contributors (I would personally use Conda environments if not using Docker, for instance).

Did you have any trouble running the project using Docker?

brunolellis:
A simple way to run the spider locally:
1. Install `requirements.txt`
2. Comment lines 7 and 8 in `settings.py` (both 'gazette.pipelines')
3. Run `scrapy crawl sp_campinas` or another spider

I don't know if it is the best way since I am not from python world, but I hope it helps :grinning:

stefersonferreira:
In windows you will need to install the pywin32.

use `pip install pypiwin32`

kerollaine:
I have no problems running the project using Docker, I just think it is easier to debug or test parts of the code without docker. 
But I agree with you, it's important to keep the project setup as simple as possible.
Thank you very much for the feedback!

Irio:
@kerollaine I believe we can have the best of both worlds if I enable the wiki and let everyone edit it so this documentation can be maintained by the community, adding a reference in the README file. How about that?

kerollaine:
@Irio the way that @brunolellis explained is simple and works very well too. So I think this is unnecessary for now, as each contributor can find their preferred way to achieve this by their own. Additionally wiki maintenance can become difficult over time, IMO.

PR #28, opened 2018-04-27
Overriding file_path

brunolellis:
As discussed in issue #15, scrapy has a bug when request URL contains query parameters.


Irio:
Awesome.

PR #29, opened 2018-04-27
Add Rio de Janeiro/RJ spider

brunolellis:
This PR creates a spider to crawl gazettes from Rio de Janeiro.

I don't know how to test all possible flows using scrapy contracts.

For example:
1. `no_gazette` for weekends and holidays
2. `one_gazette` when no extra edition is published
3. `multiple_gazettes` for extra editions

I thought about creating 3 functions, one for each case, and write the contract inside each callback function, but it seems overkill, doesn't it?

**PS: this PR depends on #28**

### DIFF
--- a/processing/data_collection/gazette/spiders/rj_rio_de_janeiro.py
+++ b/processing/data_collection/gazette/spiders/rj_rio_de_janeiro.py
@@ -58,4 +58,4 @@ class RjRioDeJaneiroSpider(scrapy.Spider):
             municipality_id=self.MUNICIPALITY_ID,
             power='executive',
             scraped_at=dt.datetime.utcnow(),
-        )        
+        )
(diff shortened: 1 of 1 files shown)


processing/data_collection/gazette/spiders/rj_rio_de_janeiro.py
```diff
@@ -0,0 +1,58 @@
...
+            for ed in editions:
+                match = re.search('.*edi_id=([0-9]+).*', ed)
+                if match:
+                    url = self.download_gazette_url.format(match.group(1))
+                    is_extra_edition = 'suplemento' in ed.lower()
+                    items.append(self.create_gazette(gazette_date, url, is_extra_edition))
+        
+        return items
+
+    def create_gazette(self, date, url, is_extra_edition = False):
```
mvinoba: Just a small remark: PEP8 recommends no spaces around = sign when indicating default paramater value.

PR #32, opened 2018-05-01
Add Maceió (AL) spider

kerollaine:
Maceió is the 17th largest city in Brazil by population. This spider collects data from http://www.maceio.al.gov.br/noticias/diario-oficial/.

The `parse_additional_page` callback is necessary because some gazettes have a direct link to download and others open a new page upon click. For the same reason the scrapy contracts have different values for min and max ranges, because in some cases we will have more or less requests per gazette. The value is unpredictable because the page contents change according to new gazette uploads.

### DIFF
--- a/processing/data_collection/gazette/spiders/al_maceio.py
+++ b/processing/data_collection/gazette/spiders/al_maceio.py
@@ -1,0 +1,76 @@
+from dateparser import parse
+from datetime import datetime
+
+import scrapy
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class AlMaceioSpider(BaseGazetteSpider):
+    MUNICIPALITY_ID = '2704302'
+    name = 'al_maceio'
+    allowed_domains = ['maceio.al.gov.br']
+    start_urls = ['http://www.maceio.al.gov.br/noticias/diario-oficial/']
+    page_number = 1
+
+    def parse(self, response):
+        """
+        @url http://www.maceio.al.gov.br/noticias/diario-oficial/
+        @returns items 0 9
+        @returns requests 1 10
+        @scrapes date file_urls is_extra_edition municipality_id power scraped_at
+        """
+        gazettes = list(response.xpath('//article'))
+        for gazette in gazettes:
+            url = gazette.xpath('a/@href').extract_first()
+
+            if not url: # In some cases the href attr is empty, e.g. 24-11-2015
+                continue
+
+            date_str = gazette.xpath('time/text()').extract_first()
+            date = parse(date_str, languages=['pt'])
+            title = gazette.xpath('a/@title').extract_first()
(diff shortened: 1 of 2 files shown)


cuducos:
Makes 100% sense ; ) Maybe we could completely remove this Scrapy contract because every 1st day of the month will break stuff…

kerollaine:
@cuducos, one another option is to define a min and max range value to the `@returns` contract - something like `@returns 0 44` -, but keep in mind that it can be hard to set a max value because we may have one gazette per workday plus extra editions. Considering 22 workdays plus 1 extra edition per day we have a 44 items limit. But I really don't now if this is better than to just remove the contract altogether.

processing/data_collection/gazette/spiders/al_maceio.py
```diff
@@ -0,0 +1,74 @@
...
+        This condition is necessary to stop crawling when date < date_limit.
+        """
+        if date >= self.date_limit:
+            self.page_number += 1
+            yield scrapy.Request(
+                '{0}/page/{1}'.format(self.start_urls[0],
+                str(self.page_number))
+            )
+
+    def  parse_additional_page(self, response):
```
rennerocha: One extra space here.

processing/data_collection/gazette/spiders/al_maceio.py
```diff
@@ -0,0 +1,74 @@
...
+                    request.meta['date'] = date
+                    request.meta['is_extra_edition'] = is_extra_edition
+                    yield request
+            else:
+                break
+
+        """
+        This condition is necessary to stop crawling when date < date_limit.
+        """
+        if date >= self.date_limit:
```
rennerocha: See https://github.com/okfn-brasil/diario-oficial/pull/41/ and https://github.com/okfn-brasil/diario-oficial/pull/46 .
There is no need to include date validation in your spider. This will be handled in an item pipeline.

processing/data_collection/gazette/spiders/al_maceio.py
```diff
@@ -0,0 +1,74 @@
...
+        @scrapes date file_urls is_extra_edition municipality_id power scraped_at
+        """
+        for gazette in response.xpath('//article'):
+            url = gazette.xpath('a/@href').extract_first()
+            date_str = gazette.xpath('time/text()').extract_first()
+            date = parse(date_str, languages=['pt'])
+            title = gazette.xpath('a/@title').extract_first()
+            is_extra_edition = 'suplemento' in (title.lower())
+
+            if date >= self.date_limit:
```
rennerocha: See #41 and #46 .
There is no need to include date validation in your spider. This will be handled in an item pipeline.

kerollaine:
Thanks a lot @rennerocha and @giovanisleite for the review. I think I've applied everything you asked.

PR #40, opened 2018-05-15
Use the correct method for retrieving the list of initial requests

rennerocha:
According to Scrapy documentation, the preferred method to return the list of the Requests that will be parsed is using the method _start_requests_:
https://doc.scrapy.org/en/latest/topics/spiders.html#scrapy.spiders.Spider.start_requests

cc/ @gustavodemari

### DIFF
--- a/processing/data_collection/gazette/spiders/sp_guarulhos.py
+++ b/processing/data_collection/gazette/spiders/sp_guarulhos.py
@@ -11,15 +11,13 @@ class SpGuarulhosSpider(scrapy.Spider):
     MUNICIPALITY_ID = '3518800'
     name = 'sp_guarulhos'
     allowed_domains = ['guarulhos.sp.gov.br']
-    urls = []
-    starting_date = dt.date(2015, 1, 1)
-    ending_date = dt.date.today()
-    for date in rrule(MONTHLY, dtstart=starting_date, until=ending_date):
-        url = "http://www.guarulhos.sp.gov.br/diario-oficial/index.php?mes={}&ano={}".format(
-            date.month, date.year
-        )
-        urls.append(url)
-    start_urls = urls
+
+    def start_requests(self):
+        base_url = 'http://www.guarulhos.sp.gov.br/diario-oficial/index.php?mes={}&ano={}'
+        starting_date = dt.date(2015, 1, 1)
+        ending_date = dt.date.today()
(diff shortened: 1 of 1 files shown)


rennerocha:
I needed to fix the contract for this spider. Guarulhos' page is returning 17 instead of 16 items. Probably a missing file was added after the contract was included to it.

Personally I don't like contracts like that, as they usually fails even when everything is correct (like this case). I prefer to check the structure of the item returned (if required fields are returned and if the format of the fields are correct) are a more useful monitoring tool.

gustavodemari:
@rennerocha I did this way (without *start_requests*) because the tutorial pointed out that it was just a shortcut to *start_requests*
https://doc.scrapy.org/en/latest/intro/tutorial.html#a-shortcut-to-the-start-requests-method

Can you explain some differences of using *start_urls* vs *start_requests*? Maybe, I may missing some point here

Also, I'm not opposed to this change, but I would like to know the differences.

Thanks

rennerocha:
The common case for using _start_urls_ list is when you have a **fixed** list of URLs you want to parse that don't requires any additional processing. For example, if we have only a fixed list of pages that we want to parse and we are sure that this will not change over time, we can use this shortcut.

When you need to do some kind of processing to generate the list of URLs before generating the requests, you should override _start_requests_ and not use _start_urls_. It is guaranteed that this method will be called only once, so it is safe to rely on it to create your requests.

Examples when you need to override: you need to log an user to get a session cookie, or you want to dynamically generate the URLs based on date (like we want to do).

If you see the code of the base _start_requests_, it just iterate over self.start_urls list and generate a simple Request:

```
    for url in self.start_urls:
        yield Request(url, dont_filter=True)
```

It doesn't mean that what you did will not work (it is working), but it is not the best way to accomplish that.

gustavodemari:
Great explanation @rennerocha 

Thanks

Irio:
@rennerocha 
> Personally I don't like contracts like that, as they usually fails even when everything is correct (like this case). I prefer to check the structure of the item returned (if required fields are returned and if the format of the fields are correct) are a more useful monitoring tool.

Interested to listen more on this topic. Contracts so far seemed a good enough and beginner-friendly way of creating integration tests. If you're up to it, let's discuss this in a dedicated issue.

PR #42, opened 2018-05-17
Curitiba

antoniovendramin:
Lendo os diários oficiais de Curitiba.
Curitiba publica os diários do executivo e legislativo no mesmo arquivo.
A plataforma onde é publicado o arquivo utiliza a tecnologia ASP.NET, com isso o site é statefull obrigando-nos a reenviar os paramentros do estado em todas as requests.

### DIFF
--- a/processing/data_collection/gazette/spiders/pr_curitiba.py
+++ b/processing/data_collection/gazette/spiders/pr_curitiba.py
@@ -10,7 +10,7 @@ from gazette.items import Gazette
 
 
 class PrCuritibaSpider(BaseGazetteSpider):
-    MUNICIPALITY_ID = '4106902'
+    TERRITORY_ID = '4106902'
     name = 'pr_curitiba'
     allowed_domains = ['legisladocexterno.curitiba.pr.gov.br']
     custom_settings = {
@@ -99,7 +99,7 @@ class PrCuritibaSpider(BaseGazetteSpider):
                     date=parsed_date,
                     file_urls=["http://legisladocexterno.curitiba.pr.gov.br/DiarioSuplementoConsultaExterna_Download.aspx?id={}".format(gazette_id)],
                     is_extra_edition=True,
-                    municipality_id=self.MUNICIPALITY_ID,
+                    territory_id=self.TERRITORY_ID,
                     power='executive_legislature',
                     scraped_at=dt.datetime.utcnow()
                 )
@@ -111,7 +111,7 @@ class PrCuritibaSpider(BaseGazetteSpider):
             date=parsed_date,
             file_urls=["http://legisladocexterno.curitiba.pr.gov.br/DiarioConsultaExterna_Download.aspx?id={}".format(gazette_id)],
             is_extra_edition=False,
-            municipality_id=self.MUNICIPALITY_ID,
+            territory_id=self.TERRITORY_ID,
             power='executive_legislature',
             scraped_at=dt.datetime.utcnow()
         )


processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,120 @@
...
+                    file_urls=["http://legisladocexterno.curitiba.pr.gov.br/DiarioSuplementoConsultaExterna_Download.aspx?id={}".format(id)],
+                    is_extra_edition= True,
+                    municipality_id=self.MUNICIPALITY_ID,
+                    power='executive_legislature',
+                    scraped_at=dt.datetime.utcnow()
+                )
+
+    def scrap_not_extra_edition(self, response):
+        parsed_date = response.meta['parsed_date']
+        id = re.findall(r'Id=(\d+)', response.text )[0]
```
rennerocha: Don't do that! You have no idea if your regex will match something. You'll get a big `IndexError` when it happens!
You can use the _re_ method in _response_. See https://doc.scrapy.org/en/latest/topics/selectors.html#using-selectors-with-regular-expressions

beothorn: The response for this request is not a valid html, I believe I can't call re from response because I can't get a valid return from any selector, correct?

About raising exceptions, if there is no "Id=" I wouldn't want to ignore this error, because it would mean that the website structure changed. I agree  that IndexError is a really bad exception to raise when this happens.
Do you have any suggestion on how should I deal with it in a way that it would be clear that something went wrong? Should I throw a more explanatory exception or is there another way to log this problem?

giovanisleite:

processing/data_collection/gazette/spiders/pr_curitiba.py
antoniovendramin: @rennerocha @cuducos @Irio We don't know what must be done when an unexpected return happen. Should we fail silently or throw an exception? We guess that is the last change before the PR can be accepted.

processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,119 @@
...
+            },
+            callback=self.parse_year
+        )
+
+
+    def parse_year(self, response):
+        for i in range(12):
+            yield self.scrap_month(response, i)
+
+    def scrap_month(self, response, month):
```
rennerocha: It is scrap**e**, not scrap.

processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,119 @@
...
+        ids = row.css("td:nth-child(3) a ::attr(data-teste)").extract()
+        for i in range(len(numbers)):
+            number = numbers[i]
+            pdf_date = pdf_dates[i]
+            id = ids[i]
+            parsed_date = parse(f'{pdf_date}', languages=['pt']).date()
+            if id == '0':
+                yield scrapy.FormRequest.from_response(
+                    response,
+                    headers = {'user-agent': 'Mozilla/5.0'},
```
rennerocha: Is it really necessary to change the `User-Agent` in this request? You are doing similar requests in other places and didn't changed it. If this is really required for this website, I suggest to set a custom user-agent for all the requests made by this spider.

Include a `custom_settings` in the spider as you can see here https://doc.scrapy.org/en/latest/topics/settings.html#settings-per-spider and set the desired `USER_AGENT` key (https://doc.scrapy.org/en/latest/topics/settings.html#user-agent)

processing/data_collection/gazette/spiders/pr_curitiba.py
rennerocha: Always try to handle the problem. But this is a avoidable problem.

In this case I didn't find where in the page this `Id=(\d+)` is.
For sure it is a string, so you can use the `re_first` method: `response.selector.re_first('Id=(\d+)')`.

processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,119 @@
...
+                    power='executive_legislature',
+                    scraped_at=dt.datetime.utcnow()
+                )
+
+    def scrap_not_extra_edition(self, response):
+        parsed_date = response.meta['parsed_date']
+        id = re.findall(r'Id=(\d+)', response.text )[0]
+        return Gazette(
+            date = parsed_date,
+            file_urls=["http://legisladocexterno.curitiba.pr.gov.br/DiarioConsultaExterna_Download.aspx?id={}".format(id)],
```
rennerocha: `id` is a built-in function in Python (https://docs.python.org/3.6/library/functions.html#id). Don't use it as a variable name. Give a better name like `pdf_id` to avoid problems.

processing/data_collection/gazette/spiders/pr_curitiba.py
rennerocha: If the page changes in future (and I hope so, because dealing with these _VIEWSTATES are terrible), we will notice monitoring it and realizing that no items were scraped.

processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,119 @@
...
+            is_extra_edition= False,
```
rennerocha: I noticed that we have gazettes like  `92 Supl 1` and `92` . But some days don't have the `Supl` version.
Isn't it to be considered as an extra edition??? @Irio @cuducos any idea?

processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,119 @@
+from dateparser import parse
+import datetime as dt
+import re
+
+import scrapy
+
+from gazette.items import Gazette
+
+class PrCuritibaSpider(scrapy.Spider):
```
rennerocha: Inherit from `BaseGazetteSpider`(https://github.com/okfn-brasil/diario-oficial/blob/master/processing/data_collection/gazette/spiders/base.py#L6) to allow date filtering by item pipelines.

processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,119 @@
...
+    MUNICIPALITY_ID = '4106902'
+    name = 'pr_curitiba'
+    allowed_domains = ['legisladocexterno.curitiba.pr.gov.br']
+    start_urls = ['http://legisladocexterno.curitiba.pr.gov.br/DiarioConsultaExterna_Pesquisa.aspx']
+
+
+    def parse(self, response):
```
rennerocha: You could put all this logic from `parse` and `scrap_year` removing the `start_urls` and implementing the `start_requests` method (https://doc.scrapy.org/en/latest/topics/spiders.html#scrapy.spiders.Spider.start_requests).

beothorn:
Implemented all suggestions and fixes
Unfortunately, we do need to set the user agent or else regular editions will not work ( yes, only regular editions 🤷  ).

processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,117 @@
...
+                'http://legisladocexterno.curitiba.pr.gov.br/DiarioConsultaExterna_Pesquisa.aspx',
+                formdata={
+                    'ctl00$cphMasterPrincipal$ddlGrAno': str(year)
+                },
+                callback=self.parse_year
+            )
+
+    def parse_year(self, response):
+        for i in range(12):
+            yield self.scrape_month(response, i)
```
rennerocha: Why create a new method that is used only here? Just yield what you are returning in L45 and avoid increasing the complexity of the spider.

processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,117 @@
...
+            yield scrapy.FormRequest(
+                'http://legisladocexterno.curitiba.pr.gov.br/DiarioConsultaExterna_Pesquisa.aspx',
+                formdata={
+                    'ctl00$cphMasterPrincipal$ddlGrAno': str(year)
+                },
+                callback=self.parse_year
+            )
+
+    def parse_year(self, response):
+        for i in range(12):
```
rennerocha: `i` is not a good variable name. It seems you are trying to search for months, right? Maybe `month` is a better name.
Anyway, this will fail in future months (for example, considering the date of this review, december/2018). Maybe include some validation to avoid trying to get future months.

processing/data_collection/gazette/spiders/pr_curitiba.py
```diff
@@ -0,0 +1,117 @@
...
+                yield Gazette(
+                    date=parsed_date,
+                    file_urls=["http://legisladocexterno.curitiba.pr.gov.br/DiarioSuplementoConsultaExterna_Download.aspx?id={}".format(gazette_id)],
+                    is_extra_edition=True,
+                    territory_id=self.TERRITORY_ID,
+                    power='executive_legislature',
+                    scraped_at=dt.datetime.utcnow()
+                )
+
+    def scrap_not_extra_edition(self, response):
```
rennerocha: `parse_regular_edition` is a better name for this method (it is scrape, not scrap anyway).

PR #43, opened 2018-05-18
Update Guarulhos and Campinas contracts

giovanisleite:
Guarulhos and Campinas added a new gazette to jan/2018 and broke our contracts.

This problem will happen again, should we avoid using an older year for the contract?


Irio:
Maybe. That's a thing: the municipalities were not supposed to be publishing gazettes in past dates.

rennerocha:
@Irio , this is the main reason that we don't usually use contracts.
They break even if the spider is not broken and most of the times, when they break it is a false positive, not a real error.

giovanisleite:
> Did you have the change of finding which days were added?

No, I was just updating my local repo, saw the error and make the pr and I don't know a way to see that (besides looking into the database if it was with these cities data). There is another way?

Irio:
@giovanisleite The only guaranteed way (at this moment) is to take a look on every pdf collected.

PR #45, opened 2018-05-19
Pr ponta grossa

antoniovendramin:
Ponta Grossa são páginas simples com a data no nome do arquivo PDF.
Legislativo e executivo são no mesmo PDF.

### DIFF
--- a/processing/data_collection/gazette/spiders/pr_ponta_grossa.py
+++ b/processing/data_collection/gazette/spiders/pr_ponta_grossa.py
@@ -6,27 +6,43 @@ import scrapy
 
 from gazette.items import Gazette
 
-
-class RsPontaGrossaSpider(scrapy.Spider):
+class PrPontaGrossaSpider(scrapy.Spider):
     MUNICIPALITY_ID = '4119905'
     name = 'pr_ponta_grossa'
     allowed_domains = ['pontagrossa.pr.gov.br']
     start_urls = ['http://www.pontagrossa.pr.gov.br/diario-oficial/']
+    ano_minimo = 2015
 
     def parse(self, response):
         """
         @url http://www.pontagrossa.pr.gov.br/diario-oficial/
         @returns requests 48
         """
-        pdf_links = response.css(".view-content .field a::attr(href)").extract()
-        for pdf in pdf_links:
-            print(pdf)
-            if "sem_atos" in pdf:
+        self.scrape_page(response)
+
+    def scrape_page(self, response):
+        pdf_links = response.css(".view-content .field a").extract()
+
+        pdf_infos = []
+        current_page = response.page-current
+        for pdf_link in pdf_links:
+            pdf_file_name = pdf_link.css("::attr(href)")
+            pdf_link_text = pdf_link.css("::text")
+            if "sem_atos" in pdf_file_name:
                 continue
-            pdf_link_info = re.search('.*/diario-oficial/_?(\d{4})-(\d{2})-(\d{2}).*.pdf', pdf)
+            pdf_link_info = re.search('.*/diario-oficial/_?(\d{4})-(\d{2})-(\d{2}).*.pdf', pdf_file_name)
             ano = pdf_link_info.group(1)
+            if ano < ano_minimo:
+                continue
             mes = pdf_link_info.group(2)
             dia = pdf_link_info.group(3)
-            print("ANO {0} MES {1} DIA {2} ".format(ano,mes,dia))
+            is_extra = True#Dapaende do texto 
+            pdf_infos.append({ "ano" : ano, "mes" : mes, "dia":dia })
+        menor_ano_da_pagina =  min(map(lambda p: p.ano, pdf_infos))
+        if menor_ano_da_pagina >= self.ano_minimo:
+            next_page_url = response.css(".pager-next::attr(href)").extract_first()
+            yield scrapy.Request(next_page_url, self.scrape_page)
 
+        for pdf_info in pdf_infos:
+            yield None
 
--- a/processing/data_collection/gazette/spiders/pr_ponta_grossa.py
+++ b/processing/data_collection/gazette/spiders/pr_ponta_grossa.py
@@ -1,5 +1,5 @@
 from dateparser import parse
-import datetime as dt
+from datetime import date, datetime
 import re
 
 import scrapy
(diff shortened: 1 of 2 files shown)


processing/data_collection/gazette/spiders/pr_ponta_grossa.py
```diff
@@ -0,0 +1,56 @@
...
+            if "sem_atos" in pdf_file_name:
+                continue
+            pdf_link_info = re.search('.*/diario-oficial/_?(\d{4})-(\d{2})-(\d{2}).*.pdf', pdf_file_name)
+            ano = int(pdf_link_info.group(1))
+            if ano < self.ano_minimo:
+                continue
+            mes = pdf_link_info.group(2)
+            dia = pdf_link_info.group(3)
+            is_extra = "complementar" in pdf_link_text
+            pdf_infos.append({ "ano" : ano, "mes" : mes, "dia":dia, "url": pdf_file_name, "is_extra_edition": is_extra })
```
cuducos: Why not extracting this bit (lines 27-40) into a custom method? This method [could be a generator](https://cuducos.me/2017/08/11/geradores-do-python-o-que-sao-e-quais-problemas-resolvem.html) (way better than appending to a list).

This would like something like:

```python
    @staticmethod
    def pdfs_info(links):
        link_pattern = '.*/diario-oficial/_?(\d{4})-(\d{2})-(\d{2}).*.pdf'
        for link in links:
            file_name = pdf_link.css('::attr(href)').extract_first()
            link_text = pdf_link.css('::text').extract_first()

            if 'sem_atos' in file_name:
                continue

            info = re.search(link_pattern, file_name)
            year = int(info.group(1))
            if year < self.ano_minimo:
                continue

            month, day = int(info.group(2)), int(info.group(3))
            yield {
                'date': date(year, month, day),
                'is_extra_edition': 'complementar' in link_text
            }
```

processing/data_collection/gazette/spiders/pr_ponta_grossa.py
```diff
@@ -0,0 +1,56 @@
...
+
+        if pdf_infos:
+            menor_ano_da_pagina =  min(map(lambda p: p["ano"], pdf_infos))
+            if menor_ano_da_pagina >= self.ano_minimo:
+                next_page_url = "{0}{1}".format("http://www.pontagrossa.pr.gov.br", response.css(".pager-next a::attr(href)").extract_first())
```
cuducos: We could increase readability here splitting the logic and using f-strings, couldn't we?

```
url_path = response.css('.pager-next a::attr(href)').extract_first()
next_page_url = f'http://www.pontagrossa.pr.gov.br{url_path}'
```

However @rennerocha might have a better strategy to get absolute URLs from relative URLs in Scrapy ; )

processing/data_collection/gazette/spiders/pr_ponta_grossa.py
```diff
@@ -0,0 +1,56 @@
...
+                yield scrapy.Request(next_page_url, self.scrape_page)
```
cuducos: If we do rename `scrape_page` as `parse` I guess we can omit the second argument of `Request` here because `self.parse` is the default (but I'm not 100% sure about it, would you mind testing it if that's the case?)

processing/data_collection/gazette/spiders/pr_ponta_grossa.py
```diff
@@ -0,0 +1,56 @@
...
+                for pdf_info in pdf_infos:
+                    date = "{0}-{1}-{2}".format(pdf_info["ano"], pdf_info["mes"], pdf_info["dia"])
```
cuducos: If `pdf_info` returns a `datetime.date` (as I suggest above) this could be simplified as:

```python
    gazette_date = pdf_info['date'].strftime('%Y-%m-%d')
```

(I just avoid using variables called `date` because if you `from datetime import date` you end up overwriting the `date` object).

processing/data_collection/gazette/spiders/pr_ponta_grossa.py
```diff
@@ -0,0 +1,56 @@
...
+                    yield Gazette(
+                            date = date,
+                            file_urls=[pdf_info["url"]],
+                            is_extra_edition=pdf_info["is_extra_edition"],
+                            municipality_id=self.MUNICIPALITY_ID,
+                            power="executive_legislature",
+                            scraped_at=dt.datetime.utcnow()
+                        )
```
cuducos: All this lines from 50 and on have an extra indentation level. You could delete 4 spaces in the beginning of each of them ; )

processing/data_collection/gazette/spiders/pr_ponta_grossa.py
rennerocha: If you have a relative URL, you should use _urlparse.urljoin_ instead of doing string manipulation.

As this is something very common to appear during the scraping, Scrapy has this method in the _response_ object:
https://doc.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Response.urljoin

```next_page_url = response.urljoin(response.css(".pager-next a::attr(href)").extract_first())```

File: processing/data_collection/gazette/spiders/pr_ponta_grossa.py (hunk 3 above)
Comment (rennerocha) [reply processing/data_collection/gazette/spiders/pr_ponta_grossa.py]: `self.parse` is the default callback for a `Request` . Get rid of `self.scrape_page` and put everything in `self.parse` :-)

File: processing/data_collection/gazette/spiders/pr_ponta_grossa.py [hunk 6]
```diff
@@ -0,0 +1,56 @@
...
+            ano = int(pdf_link_info.group(1))
+            if ano < self.ano_minimo:
+                continue
+            mes = pdf_link_info.group(2)
+            dia = pdf_link_info.group(3)
+            is_extra = "complementar" in pdf_link_text
+            pdf_infos.append({ "ano" : ano, "mes" : mes, "dia":dia, "url": pdf_file_name, "is_extra_edition": is_extra })
+
+        if pdf_infos:
+            menor_ano_da_pagina =  min(map(lambda p: p["ano"], pdf_infos))
```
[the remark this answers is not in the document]
rennerocha: Would it be better if the variables are also in english instead of portuguese?

processing/data_collection/gazette/spiders/pr_ponta_grossa.py
cuducos: `response.urljoin` for the win 🎉

giovanisleite:

antoniovendramin:
Accepted all suggestions.
Thanks!

Irio:
This municipality not only hasn't selectable text but has a printed pdf inside another.

As in the bottom left of the first page:
http://www.pontagrossa.pr.gov.br/files/diario-oficial/2018-05-01-ed2303.pdf

Irio:
Thanks for the contribution, @antoniovendramin.

PR #47, opened 2018-05-22
Salvador spider

rennerocha:
For older gazettes, they have the interesting habit to put the file from a different date.
It is also not possible to discover if it is an extra edition without parsing the gazette content. It will require a custom pipeline (specific to this city) to process the file content after it passes through _PdfParsingPipeline_ .

### DIFF
--- a/processing/data_collection/gazette/spiders/ba_salvador.py
+++ b/processing/data_collection/gazette/spiders/ba_salvador.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+import datetime
+import urllib.parse
+
+import dateparser
+import scrapy
+
+from gazette.items import Gazette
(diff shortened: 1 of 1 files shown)


rennerocha:
The date is not wrong. I discovered that some gazettes are related to more than one day (for example http://www.dom.salvador.ba.gov.br/index.php?option=com_content&view=article&id=5815:dom-7118&catid=1:dom). Updated the spider to handle this situation (unfortunately it will require an extra request for each gazette).

rennerocha:
To identify extra editions we need to parse the content of the PDF. Unfortunately, when converted, the information is not well structured, so I included a regex that for the gazettes that I saw is enough to make the classification. However it is not 100% guaranteed.

giovanisleite:

PR #49, opened 2018-05-25
Create spider to get executive gazettes from Jundiaí

rennerocha:


Irio:
It's a good candidate for the first parser of a gazette with multiple columns of text. Thanks @rennerocha.

PR #51, opened 2018-05-25
Add spider for Manaus/AM

alfakini:
Hey people,

In this PR we add the spider to collect data from [Manaus/AM gazette](http://dom.manaus.am.gov.br/diario-oficial-de-manaus/).

### DIFF
--- a/CITIES.md
+++ b/CITIES.md
@@ -13,7 +13,7 @@ Tracking of the crawlers and parsers for the top 100 cities.
 | 4 | Salvador | | | | https://github.com/okfn-brasil/diario-oficial/pull/47 |
 | 5 | Fortaleza | :soon: | | | https://github.com/okfn-brasil/diario-oficial/pull/52 |
 | 6 | Belo Horizonte | | | | https://github.com/okfn-brasil/diario-oficial/pull/33 |
-| 7 | Manaus | :soon:  | | | https://github.com/okfn-brasil/diario-oficial/pull/51 |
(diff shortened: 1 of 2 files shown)


alfakini:
@rennerocha Added the check for the extra edition. Thanks for the catch 👍

giovanisleite:

alfakini:
@giovanisleite @Irio Should we add in the same spider the executive and legislative? Seems to me that in the cases we have gazettes for each power we should add different spiders. What about `am_manaus_executive.py` and `am_manaus_legislative.py`?

giovanisleite:
I agree that we need two spiders, @alfakini, but just in case of two different websites (that is the case here) and keep only one spider for two separated powers if it is in the same site like rs_porto_alegre, what do you think?

Irio:
In a project (hopefully, in a few months) with hundreds of crawlers, would be easier to handle a maximum of one spider per municipality.

You could make them work by putting two elements in both `allowed_domains` and `start_urls`. Let me know if you need further instructions on how to do it.

rennerocha:
As both will return the same type of item `GazetteItem`, I'd suggest stop using the `start_urls` and write a `start_requests` method and then send the respective requests for each power with two separated callbacks: `parse_executive` (it will be the actual `parse`) and `parse_legislative`.

alfakini:
@Irio @giovanisleite @rennerocha Implemented the crawler for the legislative power. Thanks @rennerocha for the suggestion 👍

PR #52, opened 2018-05-25
Add spider for Fortaleza/CE

dannnylo:
In this PR we add the spider to collect data from [Fortaleza/CE gazette](http://apps.fortaleza.ce.gov.br/diariooficial/)

### DIFF
--- a/processing/data_collection/gazette/spiders/ce_fortaleza.py
+++ b/processing/data_collection/gazette/spiders/ce_fortaleza.py
@@ -1,0 +1,61 @@
+import re
+import dateparser
+
+from datetime import datetime
+from scrapy import Request, Spider
+from gazette.items import Gazette
+
+class GazetteElement:
+    MUNICIPALITY_ID = '2304400'
+    PDF_URL = 'http://apps.fortaleza.ce.gov.br/diariooficial/{}'
+
+    def __init__(self, element):
+        self.element = element
+
+    def date(self):
+        date_str = self.element.css('td:nth-child(2)::text').extract_first()
+        return dateparser.parse(date_str, languages=['pt']).date()
+
+    def url(self):
+        path = self.element.css('a::attr(href)').extract_first()
+        return self.PDF_URL.format(path)
+
+    def to_gazette(self):
+        return Gazette(
+            date=self.date(),
+            file_urls=[self.url()],
+            is_extra_edition=False,
+            municipality_id=self.MUNICIPALITY_ID,
+            power='executive_legislature',
+            scraped_at=datetime.utcnow(),
+        )
+
+
+class CeFortalezaSpider(Spider):
+    GAZETTE_ELEMENT_CSS = '.diarios-oficiais .table-responsive tbody tr'
+    NEXT_PAGE_CSS = 'ul.pagination .page-link'
+
+    allowed_domains = ['apps.fortaleza.ce.gov.br']
+    name = 'ce_fortaleza'
+
+    start_urls = []
+
+    for year in range(2015, datetime.now().year):
+        start_urls.append('http://apps.fortaleza.ce.gov.br/diariooficial/?num-diario=&content-diario=&ano-diario=' + str(year) + '&mes-diario=todos&current=1')
+
+    def parse(self, response):
+        """
+        @url http://apps.fortaleza.ce.gov.br/diariooficial/
+        @returns requests 1
+        @scrapes date file_urls is_extra_edition municipality_id power scraped_at
+        """
+
+        for element in response.css(self.GAZETTE_ELEMENT_CSS):
+           yield GazetteElement(element).to_gazette()
+
+        next_link = response.css(self.NEXT_PAGE_CSS)
+
+        if next_link:
+            page_number = next_link[-1].css('.page-link::attr(href)').extract_first().split('#')[1]
+            url = response.url.split('current')[0] + '&current=' + str(page_number)
+            yield Request(url)
--- a/processing/data_collection/gazette/spiders/ce_fortaleza.py
+++ b/processing/data_collection/gazette/spiders/ce_fortaleza.py
@@ -39,7 +39,6 @@ class CeFortalezaSpider(Spider):
             url = self.extract_url(element)
             date = self.extract_date(element)
             extra_edition = self.extract_extra_edition(element)
-            print(extra_edition)
 
             yield Gazette(
                 date=date,
@@ -50,13 +49,12 @@ class CeFortalezaSpider(Spider):
                 scraped_at=datetime.utcnow(),
             )
 
-        next_link = response.css(self.NEXT_PAGE_CSS)
+        next_link = self.extract_next_link(response)
 
         if next_link:
-            page_number = next_link[-1].css(self.NEXT_PAGE_LINK_CSS).extract_first().split('#')[1]
-            url = response.url.split('&current')[0] + '&current=' + str(page_number)
-            yield Request(url)
+            yield Request(next_link)
 
+    # Extra edition is maked with a "s" on description. Example: Diário Oficial Nº 15923s
     def extract_extra_edition(self, element):
         return element.css(self.EXTRA_CSS).extract_first()[-1] == 's'
 
@@ -67,3 +65,12 @@ class CeFortalezaSpider(Spider):
     def extract_date(self, element):
         date_str = element.css(self.DATE_CSS).extract_first()
         return dateparser.parse(date_str, languages=['pt']).date()
+
+    def extract_next_link(self, response):
+        next_link = response.css(self.NEXT_PAGE_CSS)
+        if next_link:
+            next_link_element = next_link[-1].css(self.NEXT_PAGE_LINK_CSS)
+            page_number = next_link_element.extract_first().split('#')[1]
+            return response.url.split('&current')[0] + '&current=' + str(page_number)
+
+        return false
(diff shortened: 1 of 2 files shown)


processing/data_collection/gazette/spiders/ce_fortaleza.py
```diff
@@ -0,0 +1,76 @@
+import re
+import dateparser
+
+from datetime import datetime
+from scrapy import Request, Spider
+from gazette.items import Gazette
+
+PDF_URL = 'http://apps.fortaleza.ce.gov.br/diariooficial/{}'
+
+def generate_urls_by_year():
```
rennerocha: Don't create a function to get the list of initial URLs, redefine `start_requests` method of `scrapy.Spider` class . It is the right way to do it. See https://github.com/okfn-brasil/diario-oficial/pull/40 for more information.

processing/data_collection/gazette/spiders/ce_fortaleza.py
```diff
@@ -0,0 +1,76 @@
...
+    EXTRA_CSS = 'td:nth-child(1)::text'
+    NEXT_PAGE_CSS = 'ul.pagination .page-link'
+    NEXT_PAGE_LINK_CSS = '.page-link::attr(href)'
+
+    MUNICIPALITY_ID = '2304400'
+
+    allowed_domains = ['apps.fortaleza.ce.gov.br']
+    name = 'ce_fortaleza'
+
+    start_urls = generate_urls_by_year()
```
rennerocha: Don't create a function to get the list of initial URLs, redefine start_requests method of scrapy.Spider class . It is the right way to do it. See #40 for more information.

processing/data_collection/gazette/spiders/ce_fortaleza.py
```diff
@@ -0,0 +1,76 @@
...
+            )
+
+        next_link = self.extract_next_link(response)
+
+        if next_link:
+            yield Request(next_link)
+
+    # Extra edition is maked with a "s" on description. Example: Diário Oficial Nº 15923s
+    def extract_extra_edition(self, element):
+        return element.css(self.EXTRA_CSS).extract_first()[-1] == 's'
```
rennerocha: `extract_first` may return `None` . You should define a default value if nothing is found and use the built-in string method `endswith` (https://docs.python.org/3/library/stdtypes.html) to check this field. So you will avoid raising an error.

`element.css(self.EXTRA_CSS).extract_first('').endswith('s')`

processing/data_collection/gazette/spiders/ce_fortaleza.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def extract_url(self, element):
```
rennerocha: You are getting a relative URL here. Instead of doing a string manipulation, you could use the `urljoin` method in the `Response` class (https://doc.scrapy.org/en/latest/topics/request-response.html#scrapy.http.Response.urljoin)

Also I don't understand why you create these function outside the `Spider`. You are increasing the complexity with something that could be done with just one line of code. My suggestion is to remove these functions or if you really want to use them, include them as class `methods`.

processing/data_collection/gazette/spiders/ce_fortaleza.py
```diff
@@ -0,0 +1,76 @@
...
+    def extract_date(self, element):
+        date_str = element.css(self.DATE_CSS).extract_first()
+        return dateparser.parse(date_str, languages=['pt']).date()
+
+    def extract_next_link(self, response):
+        next_link = response.css(self.NEXT_PAGE_CSS)
+        if next_link:
+            next_link_element = next_link[-1].css(self.NEXT_PAGE_LINK_CSS)
+            page_number = next_link_element.extract_first().split('#')[1]
+            return response.url.split('&current')[0] + '&current=' + str(page_number)
```
rennerocha: Instead of manipulating the URL string, try using this library and this function:
http://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.url.add_or_replace_parameter

processing/data_collection/gazette/spiders/ce_fortaleza.py
```diff
@@ -0,0 +1,76 @@
...
+    def extract_url(self, element):
+        path = element.css('a::attr(href)').extract_first()
+        return PDF_URL.format(path)
+
+    def extract_date(self, element):
+        date_str = element.css(self.DATE_CSS).extract_first()
+        return dateparser.parse(date_str, languages=['pt']).date()
+
+    def extract_next_link(self, response):
+        next_link = response.css(self.NEXT_PAGE_CSS)
```
rennerocha: This will get the next page numbers in a simpler way:
```
for page_number in response.css('ul.pagination .page-link::attr(href)').re('#(\d)+'):
    # Here you can use the http://w3lib.readthedocs.io/en/latest/w3lib.html#w3lib.url.add_or_replace_parameter
   # to update the URL and yield the request.

File: processing/data_collection/gazette/spiders/ce_fortaleza.py [hunk 7]
```diff
@@ -0,0 +1,76 @@
...
+            yield Gazette(
+                date=date,
+                file_urls=[url],
+                is_extra_edition=extra_edition,
+                municipality_id=self.MUNICIPALITY_ID,
+                power='executive',
+                scraped_at=datetime.utcnow(),
+            )
+
+        next_link = self.extract_next_link(response)
```
rennerocha: Instead of doing this and including an extra variable and an extra method, won't be easier to put only the 
`for page_number in response.css('ul.pagination .page-link::attr(href)').re('#(\d)+'):` block that I mentioned below? I don't get why do we need a method that will be used only once, and that contains 2-3 lines that are very self-explanatory on their intent.

dannnylo:
@rennerocha 
Thanks for all the tips. The code got much better now.

processing/data_collection/gazette/spiders/ce_fortaleza.py
```diff
@@ -0,0 +1,55 @@
...
+    allowed_domains = ['apps.fortaleza.ce.gov.br']
+    name = 'ce_fortaleza'
+    start_urls = ['http://apps.fortaleza.ce.gov.br/diariooficial/?mes-diario=todos']
+
+    def start_requests(self):
+        requests = []
+
+        for year in range(2015, datetime.now().year + 1):
+            year_url = w3lib.url.add_or_replace_parameter( self.start_urls[0], 'ano-diario', year)
+            requests.append(Request(year_url))
```
rennerocha: You don't need to create a list of requests. Just yield each request as soon you define it:
`yield Request(year_url)`

processing/data_collection/gazette/spiders/ce_fortaleza.py
```diff
@@ -0,0 +1,55 @@
...
+
+    allowed_domains = ['apps.fortaleza.ce.gov.br']
+    name = 'ce_fortaleza'
+    start_urls = ['http://apps.fortaleza.ce.gov.br/diariooficial/?mes-diario=todos']
+
+    def start_requests(self):
+        requests = []
+
+        for year in range(2015, datetime.now().year + 1):
+            year_url = w3lib.url.add_or_replace_parameter( self.start_urls[0], 'ano-diario', year)
```
rennerocha: `start_urls` is just a shortcut to `start_requests`. As you are using it, I think i'd be better to remove `start_urls` and create a variable `base_url` and use it:

```
def start_requests(self):
    base_url = 'http://apps.fortaleza.ce.gov.br/diariooficial/?mes-diario=todos'
    (...)
    for ...
        year_url = w3lib.url.add_or_replace_parameter(base_url, 'ano-diario', year)
```

rennerocha: Remove extra space:
`year_url = w3lib.url.add_or_replace_parameter(self.start_urls[0], 'ano-diario', year)`

processing/data_collection/gazette/spiders/ce_fortaleza.py
```diff
@@ -0,0 +1,55 @@
...
+    MUNICIPALITY_ID = '2304400'
+
+    allowed_domains = ['apps.fortaleza.ce.gov.br']
+    name = 'ce_fortaleza'
+    start_urls = ['http://apps.fortaleza.ce.gov.br/diariooficial/?mes-diario=todos']
+
+    def start_requests(self):
+        requests = []
+
+        for year in range(2015, datetime.now().year + 1):
```
rennerocha: Date filtering is done in `GazetteDateFilteringPipeline` (https://github.com/okfn-brasil/diario-oficial/blob/master/processing/data_collection/gazette/pipelines.py), so I think there is no problem in getting all gazettes available (no matter the year).

dannnylo: Hello @rennerocha 
I made the other changes.

About this comment, the gazette of Fortaleza has files of 1952,  but the server is to much unstable to download all data. 
I should get all data or I keep filtering the year? if Yes, what range is more interesting?

dannnylo:
@rennerocha 
Fixed all comments. 
Thanks.

data/territories.csv
```diff
@@ -1,5571 +0,0 @@
...
-2500775,Aparecida,PB,Paraíba
-3502606,Aparecida d'Oeste,SP,São Paulo
-5201405,Aparecida de Goiânia,GO,Goiás
-5201454,Aparecida do Rio Doce,GO,Goiás
-1701101,Aparecida do Rio Negro,TO,Tocantins
-5001003,Aparecida do Taboado,MS,Mato Grosso do Sul
-3300159,Aperibé,RJ,Rio de Janeiro
-3200508,Apiacá,ES,Espírito Santo
-5100805,Apiacás,MT,Mato Grosso
-3502705,Apiaí,SP,São Paulo
```
giovanisleite: This file shouldn't be deleted

Issue #31, opened 2018-04-30
[WIP] Spider for Ribeirão Preto/SP


samuelgrigolato:
Me and @kerollaine are working together on this.

We've managed to create a simple Postman Collection showing the sequence of requests needed to download a single day Gazette from https://www.ribeiraopreto.sp.gov.br/J015/pesquisaData.xhtml. It is available here: https://documenter.getpostman.com/view/2394724/diariooficial/RW1boKvQ.

Because the application is JSF-based, and it maintains the view data in the server, we need to do all three requests to get a PDF file.

samuelgrigolato:
Maybe it is not clear enough in the description but the `ViewState` is *not* going to be the same between different sessions, so basically this is why we need to do the initial requests, to scrap the ViewState to use in the next requests.

PR #67, opened 2018-05-31
Add Ribeirão Preto spider

samuelgrigolato:
Please check the code comments inside the commits and #31 for further info.

Worth noting that I needed to add a custom file download item pipeline similar to what @rennerocha suggested in #10. Please take a special look in this class while reviewing as it impacts every crawler (I tried to keep the default behavior though).

### DIFF
--- a/CITIES.md
+++ b/CITIES.md
@@ -36,7 +36,7 @@ The municipality id (IBGE code) can be found on [Wikipedia](https://pt.wikipedia
 | 26 | São José dos Campos | :soon: | | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/71) |
 | 27 | Osasco | | | | |
 | 28 | Jaboatão dos Guararapes | | | | |
-| 29 | Ribeirão Preto | | | [issue](https://github.com/okfn-brasil/diario-oficial/issues/31) | |
+| 29 | Ribeirão Preto | :white_check_mark: | | [issue](https://github.com/okfn-brasil/diario-oficial/issues/31) | [PR](https://github.com/okfn-brasil/diario-oficial/pull/67) |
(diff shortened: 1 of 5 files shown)


rennerocha:
I don't like the idea of using a special custom pipeline that is useful only for one spider in a way that would affect all spiders. In this case, I think to use this pipeline only for this spider is a better idea.

I did something similar in Salvador spider:
https://github.com/okfn-brasil/diario-oficial/pull/47/files#diff-7f4e787e93394a5014afdd591305acb3R21

Anyway, your pipeline seems to be something that could be used in other spiders one day, so I would rename it from `CustomFilesPipeline` to something more explanatory (as `SessionBasedFilesPipeline`, or something better because I think my idea was terrible) and leave it in the pipelines module to be easier to use it in other spiders.

samuelgrigolato:
> Anyway, your pipeline seems to be something that could be used in other spiders one day

Exactly, when designing this item pipeline I was trying to achieve a generic one, not one specific for Ribeirão Preto. The `cookiejar` + `formdata` solution probably apply to a lot of cases so it may help others, and can be further improved if needed, e.g to add headers support, without breaking compatibility with existing crawlers.

Regarding the name of the pipeline I'll try to improve it and update this PR, any suggestions would be greatly appreciated.

jaswdr:
@samuelgrigolato any plans to continue this PR?

samuelgrigolato:
Hi @jaswdr. Sometime after my last comment I decided to stop working on this and missed the opportunity and timing to properly comment and let you guys know, sorry about that. I'm still not able to resume working on this but I'm here to answer any questions one my have about the code.

rennerocha:
Closing stale PR.

PR #68, opened 2018-06-01
Add spider for Bauru/Sp

VGasparini:
Hey people,

In this PR we add the spider to collect data from [Bauru/SP gazette](http://www.bauru.sp.gov.br/juridico/diariooficial.aspx).


processing/data_collection/gazette/spiders/sp_bauru.py
```diff
@@ -0,0 +1,58 @@
...
+
+    allowed_domains = ['bauru.sp.gov.br']
+    name = 'sp_bauru'
+
+    def start_requests(self):
+        for month, year in self.MONTHS_YEARS:
+            yield Request(self.PAGE_URL.format(year, month))
+
+    def parse(self, response):
+
```
alfakini: Let's remove these two empty lines.

processing/data_collection/gazette/spiders/sp_bauru.py
```diff
@@ -0,0 +1,55 @@
...
+    PAGE_URL = "http://www.bauru.sp.gov.br/juridico/diariooficial.aspx?a={}&m={}"
+    DATE_CSS = "b::text"
+    PDF_HREF_CSS = "a::attr(href)"
+    GAZETTE_CSS = "main div.container ul ul ul li"
+
+    allowed_domains = ["bauru.sp.gov.br"]
+    name = "sp_bauru"
+
+    def start_requests(self):
+        for month, year in self.MONTHS_YEARS:
```
he7d3r: Looks like this will also request URLs for future months in the current year (e.g. December 2018). Shouldn't we avoid that?

rennerocha:
The main idea of this spider is incorrect. You are doing a brute-force test to get the gazettes, instead of looking at the gazettes pages and identify which ones are available for download.

Having this list will make your spider only work to get the gazettes between 2015 and 2018. We have older gazettes (that should be collected) and one day we will be in 2019 (and the spider will stop working correctly):
`YEARS = ["2015", "2016", "2017", "2018"]`

Having this URL as the initial URL:
`http://www.bauru.sp.gov.br/juridico/diariooficial.aspx`

For example, you can identify the years available using this expression:
`response.xpath('//a/@href[re:test(., "a=\d{4}$")]').re('a=(\d{4})')`
This line will get all <a> tags that contains in the end of the href property the pattern a=NNNN and then return the list of years available.

For each value in this list you can call "http://www.bauru.sp.gov.br/juridico/diariooficial.aspx?a={}" and open the page for each year:
```
    def parse(self, respose):
        for year in response.xpath('//a/@href[re:test(., "a=\d{4}$")]').re('a=(\d{4})'):
            yield Request(
                "http://www.bauru.sp.gov.br/juridico/diariooficial.aspx?a={}".format(year),
                callback=self.parse_year
            )
```

rennerocha:
Another way, that maybe will make things a little easier (and reduce some requests) could be requesting all the months (and the `parse_month` method will be the same as your actual `parse`:
```
    def parse(self, respose):
        for year in response.xpath('//a/@href[re:test(., "a=\d{4}$")]').re('a=(\d{4})'):
            for month in range(1,13):
                yield Request(
                    "http://www.bauru.sp.gov.br/juridico/diariooficial.aspx?a={}&m={}".format(year, month),
                    callback=self.parse_month
                )
```
Some requests may fail (if there is nothing in that month or if it a future month).

sergiomario:
This spider was completed in #196 
@VGasparini many thanks!

PR #77, opened 2018-06-03
Add spider for Contagem/MG

giovanisleite:
Header and footer of all documents states that it is about both powers
Example of extra edition url: ```http://www.contagem.mg.gov.br/arquivos/doc/4324complementar2doc-e-20180416055545.pdf```


processing/data_collection/gazette/spiders/mg_contagem.py
```diff
@@ -0,0 +1,48 @@
...
+        urls = [
+            response.urljoin(url)
+            for url in anchor_elements.css("::attr(href)").re(".+pdf")
+        ]
+        extra_editions = ["complementar" in url for url in urls]
+
+        dates_in_sentence = anchor_elements.css("p span:last-child ::text").re(
+            "(\d{1,2}\s+de\s+\w+\s+de\s+\d{4})"
+        )
+        dates = [parse(date, languages=["pt"]).date() for date in dates_in_sentence]
```
jvanz, line 31: It's not a major issue. But I would prefer import `parse` function from the `datetimeparser` with a different name. Just to not confuse it with the `parse` method from the spider. Just to improve readability.

giovanisleite: Sorry for taking all this time to fix something so simple, but now it is ready :rocket:

jvanz:
Thanks @giovanisleite !

Issue #73, opened 2018-06-02
Code formatter?


cuducos:
I see a lot of feedback here focused on code formatting. What about automatizing this? 

We could recommend [Black](https://github.com/ambv/black), and add a single line (`$ black . --check`) to our CI. This way PEP8 and other common code style linters would automatically raise a red light helping contributors to keep the code in a way that is compatible with _best practices_ among the Python community.

Also with `$ black .` we can edit the whole code base at once to fit the style guide.

Surely I'm a candidate to implement this, but I'd like to hear from you some counter-arguments ; )

cuducos:
Just out of curiosity, [this is how it would look like](https://github.com/okfn-brasil/diario-oficial/compare/cuducos-black?expand=1):

* Setting up to run b Black on the CI 27bc708 
* Adapting code base to style guide 5b300eb

samuelgrigolato:
The only problem I see with applying this right now is that the project currently have a lot of open pull requests, so some of them may (probably will) need to be refactored as well to comply to the new standards, before being ready to be merged.

Maybe this can remain open for a while, waiting until there happens to be an opportunity to apply such large scale code refactoring, i.e. the pull request backlog is shorter? Of course as the days goes by and the codebase gets larger it becomes more difficult to apply, so this needs to be considered as well.

Anyway a linter would definitely be a great addition to the project, in my opinion.

cuducos:
>  some of them may (probably will) need to be refactored as well to comply to the new standards, before being ready to be merged.

This can be done automatically with `$ docker-compose run --rm processing black .` We can ask people to run it (as we didi with `CITIES.md`) or we can run it locally when merging…

Irio:
Been personally using Black for a couple of months and don't have anything to complain about it.

If the idea is to enforce a code style, reducing the hassle of code reviews and the feedback loop, I prefer not to have to push, open a pull request and wait for the CI to find out I need to run a command and commit again. Run the command for me, which can be done using Git hooks. Check https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks and https://github.com/ambv/black#version-control-integration.

Never used this pre-commit.com (mentioned in Black's docs), but it can be achieved by adding the following to `make setup` task:

```
$ pip install black
$ printf "#!/bin/sh\n\nblack ." > .git/hooks/pre-commit
$ chmod +x .git/hooks/pre-commit
```

cuducos:
[...]

PR #84, opened 2018-06-07
Use Black code formatter

cuducos:
As discussed in #73 this PR introduces Black to auto format and check Python code style.

* 27bc708 formats the current source code with Black
* 5b300eb sets Black to run as part of the CI
* ebeb236 proposes the use of Black as a pre-commit Git hook

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ setup:
 	docker-compose pull
 	docker-compose build
 	make seed
+	pip install pre-commit
+	pre-commit install
 
 seed:
 	docker-compose up -d postgres
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ unit_test:
 
 integration_test:
 	docker-compose run --rm processing bash -c "cd data_collection && scrapy check"
+	docker-compose run --rm processing black . --check
 
 setup:
 	cp .env.example .env
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,7 @@
+repos:
+-   repo: https://github.com/ambv/black
+    rev: stable
+    hooks:
+    - id: black
+      args: [--safe]
+      language_version: python3.6
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,3 +45,16 @@ The project is backed by a test suite, which can be run with a single command.
 ```sh
 $ make test
 ```
+
+## Automated code formatting
+
+The project uses [Black](https://github.com/ambv/black) as an automated tool to format and check code style. If you ran `make setup` probably you are ready to go: a pre-commit Git hook will format code that happens to be in dissonance with the code style.
+
+Alternatively just install Black and run it by yourself:
+
+```sh
+$ pip install black
+$ black .
+```
+
+Is is important to note that the CI will fail if you commit Python code that is not in accordance with Black code style.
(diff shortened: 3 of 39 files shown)


rennerocha:
I noticed that the formatter is changing single quotes to double quotes. I like the first and if I remember well, you as well. Is it possible to not change this?

cuducos:
I do like single quotes. And my taste is very dear and special for me. In spite of that, [*special cases aren't special enough to break the rules*](https://www.python.org/dev/peps/pep-0020/#id3).

The thing is that in this case I do prefer convention over configuration. As Black default is double quote, and even if I like double quotes, I prefer keeping it simple and go with double quotes this time.

rennerocha:
@cuducos did you see this discussion about single quotes? :-)
https://github.com/ambv/black/issues/118#issuecomment-393298377

cuducos:
Just checked it. My first feeling is that such a lengthy discussion is a waste of time of developers who could be adding value to numerous projects but instead are debating about single vs. double quotes. Even worse: they are doing that in a project whose aim, in my opinion, is to clear the path from this discussion so we can focus on coding -- and not in style ; )

Also I disagree with the resolution: skipping the normalize string method actually weakens the argument of leaving style aside to focus on coding. It creates room for the resilience of such debate.

My idea in bringing Black in was to reduce amount of time we spent on style during our code reviews. That was my purpose, and I'm not sure it is reflected in everyone else's mind. Thus I stick to this point: I'll go with whatever Black suggests in its default. This sounds like the shortest path to keep the focus on coding, not in style.

Skipping normalization of strings would make room for code reviews such as _you're inconsistent in the use of quotes, using single quotes here and double quotes elsewhere_. I would like people from this community here to invest their valuable time in more valuable contributions.

I'm really sorry if my arguments sounds aggressive. But the point is I'm still not convinced to consider this case a special case ; ) And surely this is open to debate and I'm happy to be downvoted!

rennerocha:
You weren't aggressive at all :-)

I see some good points for using single-quotes as summarized [here](https://github.com/ambv/black/issues/118#issuecomment-392311618) and other good points for using double-quotes. As both ways are good (and bad), and this project never had a standard (we have a mix of both ways), set the standard to double-quotes is not that bad anyway. When this will be merged? :-)

I liked the solution they implemented, leaving it as optional. I don't think it weakened the argument of leaving style aside to focus on coding. Most of the resolutions are not being questioned like this one, so to me this scenario is the same we face when we chose to have a line with length bigger than 80 chars. I don't see any other rule that will promote a big debate like this.

Makefile
```diff
@@ -5,6 +5,7 @@ unit_test:
 
 integration_test:
 	docker-compose run --rm processing bash -c "cd data_collection && scrapy check"
+	docker-compose run --rm processing black . --check
```
Irio: If the intention is to run this command in the continuous integration, it should be placed in the `travis.yml` file. Travis CI actually [runs](https://github.com/okfn-brasil/diario-oficial/blob/3741d244f0496bcacb4d6b1d8259917c65a9cc73/.travis.yml#L16-L17) `make unit_test` after setup.

cuducos: Hum… ok. Semantically it is not an integration test. I just wanted to find a way to offer a kind of shortcut for local checks. Should we add another command to `Makefile` in this case (in addition to adding this command to `.travis.yml`)?

Irio: Since you added pre-commit installation to the Makefile, you could create a new `make ci` command to run the existing script section plus black's check.

cuducos: > Since you added pre-commit installation to the Makefile, you could create a new `make ci` command to run the existing script section plus black's check.

Sounds good! Gonna register this on an issue and get back to it one of these days : )

PR #89, opened 2018-06-09
Fix permission denied errors on running from scratch

viniciusartur:
This PR fixes "Permission denied" errors on starting the processing component of Diário Oficial project.

The errors are:
- Given we cloned the repository as a normal user Linux with id 1000
- When we run 'make setup && docker-compose up'
- Then the processing component gives ' _gdbm.error: [Errno 13] Permission denied' creating the schedule database of Celery Beat.
- And after fixing this error
- Then we have a Permission denied error creating the data directory of the processing task

After investigating the issue, I concluded the system user gazette (id 101) has no permissions to write in a directory 1000:1000.
For the first error, the solution I've found is just pointing the schedule database file to be created in a directory with enough permissions.
For the second error, the  solution I've found is creating a docker volume for the directory that stores the PDFs and text files after processing the spiders.


processing/data_collection/gazette/settings.py
```diff
@@ -9,4 +9,4 @@
     'gazette.pipelines.PdfParsingPipeline': 200,
     'gazette.pipelines.PostgreSQLPipeline': 300,
 }
-FILES_STORE = '/mnt/data/'
+FILES_STORE = '/mnt/files/'
```
Irio: Why move it to a different folder?

viniciusartur: Because having only /mnt/data storing territories.csv and full directory (that contains PDFs and text files) is not possible, due to permission errors.
If we create the volume to /mnt/data, there's no territories.csv available to the stage 'make setup'.
Creating a separate volume to store the PDFs and text files, we still have territories.csv available to populating the database e files scraped without permission errors.

viniciusartur:
I think there's a way to simulate the permission denied error.
What I've been doing is changing the crontab to 5 minutes + 3 hours from now.
I will explain with more details:

(Supposing now is 19:45)
1.  docker-compose down (to remove the containers)
2. change line 23 of processing/tasks.py to crontab(hour=22, minute=50)
3. docker-compose up

When the containers are running, if you check date in the processing containers, it will be around 22:45. Then at 22:50 the celery will trigger the task that will try to create the data/full directory.

In my case, my linux user has id 1000, and gazette has id 102. So when the user with id 102 doesn't have permissions to write in a directory whose owner has id 1000.

processing/data_collection/gazette/settings.py
Irio: > If we create the volume to /mnt/data, there's no territories.csv available to the stage 'make setup'.
You can you the [COPY instruction](https://docs.docker.com/engine/reference/builder/#copy) to make it available.

viniciusartur: Good point! I'm refactoring...

Irio:
Great @viniciusartur. Now I was able to reproduce it and everything seems to be back in place.

Thanks!

PR #90, opened 2018-06-13
Add spider for Blumenau/SC

goblinbr:


processing/data_collection/gazette/spiders/sc_blumenau.py
```diff
@@ -0,0 +1,56 @@
...
+    def parse(self, response):
+        """
+        @url https://www.diariomunicipal.sc.gov.br/site/?r=site/index&q=cod_municipio%3A164
+        @returns requests 1
+        @scrapes date file_urls is_extra_edition territory_id power scraped_at
+        """
+
+        possible_gazettes = response.css(self.POSSIBLE_GAZETTE_CSS);
+        for element in possible_gazettes:
+            url = self.extract_url(element)
```
rennerocha: Why create a class method that is used only once and have only one line?
`url = element.css('a::attr(href)').extract_first()` is clear enough on its intent and make it easier to maintain later (don't need to go over the code to understand what is happening)

processing/data_collection/gazette/spiders/sc_blumenau.py
```diff
@@ -0,0 +1,56 @@
...
+                yield Gazette(
+                    date=date,
+                    file_urls=[url],
+                    is_extra_edition=False,
+                    territory_id=self.TERRITORY_ID,
+                    power='executive_legislature',
+                    scraped_at=datetime.utcnow(),
+                )
+        
+        next_page_url = self.extract_next_page_url(response)
```
rennerocha: Removing the extra method will reduce the number of lines and make it easier to understand the code.
```
    next_page_path = response.css(self.NEXT_PAGE_CSS).extract_first()
    if next_page_path:
        yield response.urljoin(next_page_path)
```

When you receive a relative URL (for example "/site/?r=site/index&amp;q=cod_municipio%3A164&amp;Search_page=2"), you don't need to do string concatenation. Use the `response.urljoin` method available in Scrapy.

So replace `yield self.NEXT_PAGE_URL.format(next_page_path)` to `yield response.urljoin(next_page_path)`

goblinbr: Didn't know urljoin. I will make the changes, thanks

jvanz:
@sergiomario , can we close this one? Blumenau has been added in the PR https://github.com/okfn-brasil/diario-oficial/pull/141

PR #93, opened 2018-07-07
Don't expose Postgres port in Docker

vitorbaptista:
This avoids `docker-compose up` failing if ran on a machine that has Postgres
running locally on the same port. As the other containers access Postgres via
Docker, they'll continue working regardless if we're exposing its port or not.
Also, if we need to access the DB for whatever reason, we can get the random IP
assigned by Docker using `docker inspect` and use it instead.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,8 +15,6 @@ services:
     env_file:
       - .env
     image: postgres:10-alpine
-    ports:
-      - 5432:5432
 
   processing:
     build: processing


cuducos:
Usually we keep the port exposed because some users have database tools (PG Admin, Psequel, pghero) running locally and like to explore the database with their tools. What do you think, @irio?

vitorbaptista:
Just to clarify, the options are exposing the 5432 port, which has the problem that it won't work if you already have postgresql running locally, or not exposing it and accessing the DB container IP directly if you want to use `PG Admin` or other tools (using `docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' diariooficial_postgres_1`)

Irio:
Good to merge. Thanks, @vitorbaptista.

PR #98, opened 2018-07-20
Add spider for Go Aparecida de Goiânia

weibemoura:


processing/data_collection/gazette/spiders/go_aparecida_de_goiania.py
```diff
@@ -0,0 +1,37 @@
+import datetime as dt
+from json import loads as json_load
```
rennerocha: This is not a good practice for me. What the problem with using `json.loads()`?
When I see `json_load` in other part of the code, it is not explicit that this is the function from the standard library. My guess would be that you created a custom function to deserialize a JSON string.

`import json` and then use `json.loads()` when you need it.

processing/data_collection/gazette/spiders/go_aparecida_de_goiania.py
```diff
@@ -0,0 +1,37 @@
...
+        download_url = 'https://webio.aparecida.go.gov.br/diariooficial/download/{}'
+
+        items = []
+        records = json_load(response.text)['records']
+        for record in records:
+            url = download_url.format(record['numero'])
+            power = 'executive_legislature'
+            date = parse(record['publicado'], languages=["en"]).date()
+
+            items.append(
```
rennerocha: You don't need to create a list with all the gazettes and then return them. Simply yield the items as they are being generated. Use the following and remove the `items` list from the code:
```
yield Gazette(
    date=date,
    file_urls=[url],
    is_extra_edition=False,
    territory_id=self.TERRITORY_ID,
    power=power,
    scraped_at=dt.datetime.utcnow(),
)) 
```

processing/data_collection/gazette/spiders/go_aparecida_de_goiania.py
```diff
@@ -0,0 +1,37 @@
...
+                Gazette(
+                    date=date,
+                    file_urls=[url],
+                    is_extra_edition=False,
+                    territory_id=self.TERRITORY_ID,
+                    power=power,
+                    scraped_at=dt.datetime.utcnow(),
+                ))
+
+        return items
```
rennerocha: Remove this as you will be yielding the items inside the loop.

processing/data_collection/gazette/spiders/go_aparecida_de_goiania.py
```diff
@@ -0,0 +1,37 @@
...
+class GoAparecidaDeGoianiaSpider(BaseGazetteSpider):
+    TERRITORY_ID = "5201405"
+    name = "go_aparecida_de_goiania"
+    allowed_domains = ["aparecida.go.gov.br"]
+    start_urls = ["https://webio.aparecida.go.gov.br/api/diof/lista"]
+
+    def parse(self, response):
+        download_url = 'https://webio.aparecida.go.gov.br/diariooficial/download/{}'
+
+        items = []
```
rennerocha: Remove this as you will be yielding the items inside the loop. (see next comment).

processing/data_collection/gazette/spiders/go_aparecida_de_goiania.py
```diff
@@ -0,0 +1,37 @@
...
+        records = json_load(response.text)['records']
```
rennerocha: `records = json.loads(response.text)['records']` is much clearer and explicit!

weibemoura:
Thanks @rennerocha I just updated the code

PR #100, opened 2018-08-01
Add spider for Vitoria da Conquista/BA

he7d3r:


processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
```
cuducos: No need for that as we use Python 3 I guess.

processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        @url http://dom.pmvc.ba.gov.br/
+        @returns items 1
+        @scrapes date file_urls is_extra_edition territory_id power scraped_at
+        """
+        menu_xpath = '/html/body/div[2]/div[4]/div/div[1]'
+        menu = response.xpath(menu_xpath)
+        years_xpath = 'div[1]/ul/li/a/text()'
+        links_xpath = '//a[text()[normalize-space(.) = "Visualizar Diário"]]/@href'
+        for y in menu.xpath(years_xpath).extract():
```
cuducos: What about `y` instead of `year`?
And where do you use this variable in the code?

processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
```diff
@@ -0,0 +1,35 @@
...
+            for m in range(1, 13):
```
cuducos: What about `m` instead of `month`?
And where do you use this variable in the code?

processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
```diff
@@ -0,0 +1,35 @@
...
+                urls = response.xpath(links_xpath).extract()
+                for url in urls:
+                    yield Gazette(
+                        date=parsing_date,
```
cuducos: Where does this variabla `parsing_date` comes from?

processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
```diff
@@ -0,0 +1,35 @@
...
+        @returns items 1
+        @scrapes date file_urls is_extra_edition territory_id power scraped_at
+        """
+        menu_xpath = '/html/body/div[2]/div[4]/div/div[1]'
+        menu = response.xpath(menu_xpath)
+        years_xpath = 'div[1]/ul/li/a/text()'
+        links_xpath = '//a[text()[normalize-space(.) = "Visualizar Diário"]]/@href'
+        for y in menu.xpath(years_xpath).extract():
+            for m in range(1, 13):
+                urls = response.xpath(links_xpath).extract()
```
cuducos: Wouldn't this be the URL for a page with the links to the gazettes and not the PDF address?

processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
```diff
@@ -0,0 +1,35 @@
...
+    allowed_domains = ['pmvc.ba.gov.br']
+    start_urls = ['http://dom.pmvc.ba.gov.br/']
+
+    def parse(self, response):
+        """
+        @url http://dom.pmvc.ba.gov.br/
+        @returns items 1
+        @scrapes date file_urls is_extra_edition territory_id power scraped_at
+        """
+        menu_xpath = '/html/body/div[2]/div[4]/div/div[1]'
```
rennerocha: It is not a good practice to create xpaths with that granularity. For example, to get the years that are available in this page you can use:
```
available_years = response.css('div.visible-xs ul.dropdown-menu li a::text').re('(\d{4})')
```

processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
rennerocha: `y` is a bad variable name. You could use:
```
for year in available_years:
    # do what you want to do
```

processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
rennerocha: `m` is a bad variable name. You could use:
```
for month in range(1, 13):
    # Do what you need to do
```

processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
rennerocha: For each month, you have access to an URL like:
`http://dom.pmvc.ba.gov.br/diarios/{year}/{month}`
You need to send a request to this URL and then process the result in a new parse method.

processing/data_collection/gazette/spiders/ba_vitoria_da_conquista.py
rennerocha: `parsing_date`is wrong here. BTW, this will never work, because it is an undefined variable. Did you execute this job at least once????

`response.css('.box-diario')` contains the info you need to work for each gazette.

he7d3r:
recheck

PR #106, opened 2018-08-26
WIP: Add parser for Feira de Santana/BA

pauloabelha:
I am working on this code with @vitorbaptista 

## TODO

* Validate existence of required fields (e.g. "NUMERO")
* Decide what to do with "Extratos de contrato". Are they bidding exemptions? For example, check 1YTABV622018.pdf


vitorbaptista:
@cuducos As you can see from https://github.com/okfn-brasil/diario-oficial/pull/106#issue-210968429, the PDFs from Feira de Santana have many different formats. The code can handle some of the formats, but not all. Should we wait until we're able to extract everything, or is it better to send a series of smaller PRs?

Say we're able to extract all data from 2018, and the code is intelligent enough not to break on previous years (or extract incorrect data). Is it better to send a PR for 2018 and work on the previous years on a seprate PR, or wait until everything is ready?

(As a separate question, how important is it to extract old data at all?)

pauloabelha:
I strongly suggest that we first try to make the code work for 2018 PDFs and that working for 2018 should be good enough for a PR. A reasonable approach is to define 10% of PDFs as a test set and if the code is able to extract correctly for 10% we assume extrapolation for the other 90%. We have been defining JSONs that contain the extracted data so we can test.

As a first suggestion, to start our discussion, recommend the following steps (that might be useful for other people working on parsers):
1) Create a test set of JSONs with extracted data for 10% of all available PDFs for 2018.
2) Write the test code for comparing extracted data with JSONs
3) If code passes usual tests of not braking and extracts correctly for all 10% PDFs for extraction, we consider it good enough to be submitted as a pull request

We can tweak the percentage of the test set. Of course,a percentage of 100% just means we are extracting the data by hand :)

anapaulagomes:
Hi guys! Is this "todo" updated? I can see you've done a great work here. I'd be interested in helping to finish it. :)

vitorbaptista:
Hi @anapaulagomes, sorry for the late reply and thanks for your interest. Apart from the TODOs in @pauloabelha's first message, we just need to ensure at least the 2018 PDFs are extracted successfully. Maybe they already are. To test, I would:

1. Run the scrapers on all 2018 PDFs (there should be some docs on how to run the scraper, though I'm not sure)
1. Pick a few random examples (like the first and last DOs of 2018) and see if the extracted data is correct
1. If yes, great! Maybe leave the scraper running overnight just to see if it throws some error?
1. If not, add a new fixture on `processing/tests/gazette/fixtures/ba_feira_de_santana/` and fix the error

Also, let us know if you have any suggestion or question to improve the code.

processing/gazette/locations/ba_feira_de_santana.py
```diff
@@ -0,0 +1,139 @@
+import re
+import string
+import datetime
+
+from .base_parser import BaseParser
```
anapaulagomes: I couldn't find this module neither here or in master. Am I missing something?

processing/tests/gazette/locations/test_ba_feira_de_santana.py
```diff
@@ -0,0 +1,193 @@
+import datetime
+import json
+import pytest
```
anapaulagomes: We have to add `pytest` to `requirements.txt`. :sweat_smile:

processing/tests/gazette/locations/test_ba_feira_de_santana.py
```diff
@@ -0,0 +1,193 @@
...
+
+from gazette.locations.ba_feira_de_santana import BaFeiraDeSantana
```
anapaulagomes: I wasn't able to run the tests because the imports weren't found. By the way: what's the right way of running the tests here? Just hitting `pytest`? Would be nice to add it to `test` on Makefile.

processing/tests/gazette/locations/test_ba_feira_de_santana.py
```diff
@@ -0,0 +1,193 @@
...
+
+class TestBaFeiraDeSantana1CN1G31852018(SampleGazetteScenario):
+    GAZETTE_ID = "1CN1G31852018"
+
+
+class TestBaFeiraDeSantana1CQN162852018(SampleGazetteScenario):
+    GAZETTE_ID = "1CQN162852018"
+
+
+@pytest.mark.skip()
```
anapaulagomes: Are we skipping them because something has to be adapted?

vitorbaptista:
I'm closing this, as there are no more parsers in this repository. The code can be reused when we start parsing again. I created an issue on https://github.com/okfn-brasil/querido-diario-data-processing/issues/2 mentioning this.

Issue #108, opened 2018-08-28
Should we extract "Inexigibilidade de licitação" as well?


vitorbaptista:
While working on #106, I found a few "inexigibilidade de licitação", which seems to be cases where there is only one supplier (e.g. artists for concerts). For example, see page 5 on http://www.diariooficial.feiradesantana.ba.gov.br/atos/executivo/12L2UO1692015.pdf. Would we like to extract it as well?

giovanisleite:
I had the same question, it was answered [here](https://github.com/okfn-brasil/diario-oficial/issues/36).
I think this type of documents could bring problems not just related to value, but with the type of service as well (when we know that there is more than one supplier).

vitorbaptista:
@giovanisleite Thanks for the link! So, bottom line is: yes? If so, maybe we should add this information to the `CONTRIBUTING.md` file?

cuducos:
> Thanks for the link! So, bottom line is: yes?

Yep : )

Thanks @vitorbaptista and @giovanisleite

vitorbaptista:
Great! Sent a small PR to add this information to CONTRIBUTING.md.

PR #110, opened 2018-08-28
Explain which gazettes parts we want to extract

vitorbaptista:
As per #108, we want both bidding exemptions and ineligibilities.


CONTRIBUTING.md
```diff
@@ -38,6 +38,8 @@ Since the goal is to have a hundred municipalities into the system, it is reason
 
 Won't be perfect or as generic as we would like but maybe enough to populate a database to train Machine Learning and Natural Language Processing solutions. Meaning: the code you will find - and write - in the [`processing/gazette`](processing/gazette) folder is under heavy experimentation. Doesn't need to be beautiful or scalable if it works for specific municipalities.
 
+We are interested only in the bidding exemptions (dispensas de licitação) and bidding ineligibilities (inexigibilidade de licitação).
```
giovanisleite: It is great, to be more precise, I would add "**For now**, we are (...)".

vitorbaptista: Good point. I just added your suggestion.

Issue #109, opened 2018-08-28
Suggestion: Use pytest and tox


vitorbaptista:
I'd like to suggest using [pytest](https://pytest.org) instead of unittest. It's a very small change, the current tests don't need to change at all, and for it we'd get a few nice features from pytest. The one I'm missing the most are more useful assert error messages. For example, if I do `assert my_dict == other_dict` with unittest and it fails, I just get `AssertionError` thrown, without any message. If I use the same code with pytest I get a helpful message saying which elements of the dicts aren't the same. The same for comparing arrays, dicts, or any other element types.

And tox is useful to allow the tests to run using a single command, instead of having to remember `python -m unittest discover` when I want to run the tests outside of Docker. It keeps the virtualenvs configured, sets whatever env variables are necessary, and can also run lint.

These 2 libraries became the standard for most Python projects I see, and for good reason: they're great!

I'd be happy to send a PR with these changes (or just pytest) if you agree.

cc @irio @cuducos

cuducos:
Just to clarify: I do like `pytest` and have nothing against bringing it in. My doubt is about `tox`: if we have `pytest` we can run tests with `$ pytest`, so there's no `$ python -m unittest discover`. This checks out the _run using a single command_, right? Thus can you clarify what you mean by _it keeps the virtualenvs configured, sets whatever env variables are necessary, and can also run lint_?

I mean… virtualenv and envvars should be configured in order to run the project, so I'm not sure about your point about _fixing_ that using `tox`… I'm afraid that encapsulating it for tests can pass the wrong impression that the dev environment is OK when it's not.

vitorbaptista:
@cuducos `tox` is useful because we have multiple test types: unit, integration and linting. Personally, I like `tox` to run all the tests (including linting), and allow things like `tox -e unit`, `tox -e integration` and `tox -e lint` for when you want to run each one separately.

The virtualenv and envvars should be configured in order to run the project, but the env vars for test and for development can (and usually are) different, for example `NODE_ENV=development` vs `NODE_ENV=test`, switching the test and development databases being used, etc. Same for keeping the test and development dependencies separated.

cuducos:
> @cuducos `tox` is useful because we have multiple test types: unit, integration and linting. 

Right now we don't have unit tests separated from integration tests, and linting is in a pre-commit Git hook.

>  the env vars for test and for development can (and usually are) different

I can't see a use case for that in this project. Would you mind expanding on it?

[...]

PR #111, opened 2018-08-30
Use pytest

vitorbaptista:
As we chatted in #109, this PR migrates the code from unittest from pytest.

On tox, I was able to set it up, but there were many issues because the `gazette` user inside the `processing` container can't write to the `/mnt/code` volume, as it comes from the host. I found two workarounds:

1) On the `ENTRYPOINT` use the script `docker-entrypoint.sh` (below) to change the `gazette`'s user ID to the owner of the `/mnt/code` (I can't simply chown the folder because that'll alter the owner in the user's host machine). This pattern is relatively common in Dockerland.
2) Configure `tox` and `pytest` to write their files outside `/mnt/code` (e.g. in `/mnt/data/.tox`)

Both work fine, although the 1st option caused other issues because we're sharing the `.pyc` files amongst different "machines" (the host and the guest). I got this exception:

```
import file mismatch:
imported module 'tests.gazette.locations.test_rs_porto_alegre' has this __file__ attribute:
  /home/vitor/Projetos/diario-oficial/processing/tests/gazette/locations/test_rs_porto_alegre.py
which is not the same as the test file we want to collect:
  /mnt/code/tests/gazette/locations/test_rs_porto_alegre.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

So the 2nd option was the best. But then I noticed the `scrapy check` tests don't run locally anyway (they hardcode paths like `/mnt/code`, and require the database to be up and running), so tox was only used to take care of the virtual envs and as a single place to document how to run the tests. The virtual envs aren't necessary inside the Docker container, and the Makefile is already enough documentation IMO.

Because of this, I think tox would only be useful if we want to move the linting out of a git hook into it (which I slightly prefer). As this turned out to be more complex and less useful than I anticipated, I went on and only migrated to pytest for now. I'll add the `tox.ini` file below, so we can use it if you'd like to migrate. Otherwise, this PR is good to go.

## Code to add tox

`processing/docker-entrypoint.sh`
```bash
#!/bin/bash

OWNER_UID=$(stat -c '%u' /mnt/code)
USER=gazette

usermod -u $OWNER_UID $USER
chown -R $USER /mnt/data

gosu $USER $@
```

`processing/tox.ini`
```
[tox]
envlist =
  py36
skipsdist = true

[testenv]
passenv =
  API_URL
  DATABASE_URL
  DEBUG
  PARSING_FREQUENCY_IN_SECONDS
  PGRST_DB_ANON_ROLE
  PGRST_DB_SCHEMA
  POSTGRES_PASSWORD
  POSTGRES_USER
  RABBITMQ_URL
  REDIS_URL
deps =
  -rrequirements.txt
  freezegun==0.3.10
  pytest
commands = pytest {posargs}

[testenv:scrapy]
changedir = {toxinidir}/data_collection
setenv =
  PYTHONPATH = {toxinidir}/data_collection{:}{toxinidir}
commands = scrapy check {posargs}
```

`Makefile`
```
unit_test:
	docker-compose run --rm processing tox --workdir /mnt/data/.tox -- -p no:cacheprovider

integration_test:
	docker-compose run --rm processing tox -e scrapy --workdir /mnt/data/.tox
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ htmlcov/
 .coverage
 .coverage.*
 .cache
+.pytest_cache
 nosetests.xml
 coverage.xml
 *.cover
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 test: unit_test integration_test
 
 unit_test:
-	docker-compose run --rm processing python -m unittest discover
+	docker-compose run --rm processing pytest -p no:cacheprovider
 
 integration_test:
 	docker-compose run --rm processing bash -c "cd data_collection && scrapy check"
--- a/processing/data_collection/gazette/middlewares.py
+++ b/processing/data_collection/gazette/middlewares.py
@@ -6,7 +6,7 @@
 from scrapy import signals
 
 
-class GazetteSpiderMiddleware(object):
+class GazetteSpiderMiddleware:
 
     # Not all methods need to be defined. If a method is not defined,
     # scrapy acts as if the spider middleware does not modify the
@@ -50,7 +50,7 @@ class GazetteSpiderMiddleware(object):
         spider.logger.info("Spider opened: %s" % spider.name)
 
 
-class GazetteDownloaderMiddleware(object):
+class GazetteDownloaderMiddleware:
 
     # Not all methods need to be defined. If a method is not defined,
     # scrapy acts as if the downloader middleware does not modify the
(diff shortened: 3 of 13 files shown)


vitorbaptista:
@cuducos Oops! I guess I'm still not in a world with Py3 only. Old habits die hard :sweat_smile: 

I removed the `(object)` and fixed the code linting. It should be good to go now.

PR #119, opened 2018-10-02
Fix pre-commit installation in the setup

cuchi:
**PIP** defaults the the package installations to a system directory, which requires root access. This is not desirable while running a setup script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,11 @@ setup:
 	docker-compose pull
 	docker-compose build
 	make seed
-	pip install --user pre-commit
+	@if [ -z $$VIRTUAL_ENV ]; then \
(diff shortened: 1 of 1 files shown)


cuducos:
Hi @cuchi. I agree with your argument but it looks like it's not _always_ the case. For instance your proposed change broke the CI: `Can not perform a '--user' install. User site-packages are not visible in this virtualenv.`

Any ideias anyone?

cuducos:
Would you mind rebasing from `master` to make sure the CI passes?

Irio:
The CI passes through this specific step. I will merge it and check how it goes with master. Thanks @cuchi and @cuducos.

PR #123, opened 2018-10-15
Simplify web container

cuducos:
This commit simplifies the web container

* by avoiding installing npm packages globally (we can call them using npx or via npm scripts
instead)
* by using the prestart npm script to avoid the need of chaining two commands to run the server

Does that make sense, @guidiego?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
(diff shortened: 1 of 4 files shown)


guidiego:
Yeah! That gives more control to NPM itself :) It's a good practice avoid global dependencies o/

PR #125, opened 2018-10-22
Use latest black version

### LINKED ISSUE #122: Black pre-commit unstable
Back pre-commit it is unstable.

When I run `docker-compose run --rm processing black . --check`, I get
```
(...)
would reformat /mnt/code/tests/gazette/data/test_bidding_exemption_parsing.py
All done! ✨ 🍰 ✨
1 file would be reformatted, 56 files would be left unchanged.
```
So I run `docker-compose run --rm processing black . --safe` and this file is formatted.

But when I try to commit, pre-commit prevent and format the file back to the format that black from docker complained about it (test_bidding_exemption_parsing).

I suspect that the black inside the docker and the black used in the pre-commit are different, but I do not know how to turn them equal

----------
I already tried to: 
- Install black globally in the same version of requirements.txt.
- Use the `entry` argument on `.pre-commit-config.yaml` using the docker command of black.

giovanisleite:
Fix #122

### DIFF
--- a/processing/requirements.txt
+++ b/processing/requirements.txt
@@ -1,4 +1,4 @@
-black==18.5b1
+black==18.9b0
 celery==4.1.1
 dateparser==0.7.0
 freezegun==0.3.10
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/ambv/black/
-    rev: 18.5b1
+    rev: 18.9b0
     hooks:
     - id: black
       args: [--safe]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
--   repo: https://github.com/ambv/black
-    rev: stable
+-   repo: https://github.com/ambv/black/
+    rev: 18.5b1
     hooks:
     - id: black
       args: [--safe]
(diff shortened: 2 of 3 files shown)


processing/requirements.txt
```diff
@@ -1,4 +1,4 @@
-black==18.5b1
+black
```
cuchi: I would use fixed versions for that kind of dependency.
```suggestion
black==18.9b0
```

giovanisleite: With fixed version, we break the pre-commit. We need the pre-commit with a fixed version (but reading [here](https://github.com/ambv/black#version-control-integration), I understood that we have only two options. Use the last stable version (my change) or the master branch (it is not what we want.)
black is a dependency about code-style, so it don't affect the operation, just the code itself.

giovanisleite: The way that it is today (fixed version), we will have problems like #122. Local black considers the code ok, pre-commit doesn't and fix it... forever.

cuducos: Isn't it an option to fix `black` and `pre-commit` versions to something that actually work together?

giovanisleite: I thought that there was just two options to rev: `stable` or `master`, but it worked.

giovanisleite:
@cuchi @cuducos It looks like they're agreeing with each other now

Issue #44, opened 2018-05-18
Replacement for Scrapy Contracts


Irio:
@rennerocha recently [raised the issue](https://github.com/okfn-brasil/diario-oficial/pull/43#issuecomment-390254091) of Scrapy Contracts often making the test suite fail due to false positives.

The initial idea of using Contracts was to have a quick (for experienced developers) and beginner-friendly way of writing Scrapy tests, exempting developers from having to write test cases, creating fixtures and maintaining them up-to-date.

We could also keep using Contracts, knowing they will eventually fail in online tests, using them only as a way of monitoring the state of all crawlers (today we have 13 different spiders) until we have a better and more robust one.

cuducos:
I think that tests that depend on external resources (an external URL, the DOM of an external webpage etc.) is always troublesome when we try to make it part of our CI. The suite will eventually fail because of something nothing related to our code. However, this is gold IMHO:

> The initial idea of using Contracts was to have a quick (for experienced developers) and beginner-friendly way of writing Scrapy tests, exempting developers from having to write test cases, creating fixtures and maintaining them up-to-date.

I think we could be less pragmatic about Scrapy Contracts: we know they will fail. We could use them not as part of the CI, but as part of an sort of _integration monitor_ outside the CI.

This will avoid raising red flags in PRs (that might not be friendly for beginners), and make the PR/review/merge process more seamless. But we keep it running elsewhere and it will let us know when some Spider needs to be updated, or even tell us stories about cities publishing gazettes in months that are way past behind…

rennerocha:
I have some discussions at the company on better ways to test spiders. One thing that we mostly agreed is that it is better to monitor the spider execution than try to unit test its code, as the pages changes and the content is not always reliable. We even have a tool (unfortunately not open-sourced yet) that we can include monitoring (checking the existence of errors, the structure of the data returned, etc). It is not ideal so if someone else have better ideas it will be great to hear!

One thing that could be tested are utilitarian code (like pipelines, processors, simple functions) that we use and it is not related to the content of the pages we are scraping. We could think about custom contracts (https://doc.scrapy.org/en/latest/topics/contracts.html#custom-contracts) to create better ways to validate if the spider is working well, but I have never used it.

But I agree with @cuducos that it'd be better if contracts failures didn't break the build, so we won't have a lot of red flags in the PRs.

Irio:
Just changed the CI to run only unit tests, leaving the contracts for when running in the local environments.

[...]

PR #126, opened 2018-11-06
Fixing integration test sp_guarulhos

rqclima:
sp_guarulhos integration test was failing, it seems that the available gazettes in that month had changed from 17 to 16.

### DIFF
--- a/processing/data_collection/gazette/spiders/sp_guarulhos.py
+++ b/processing/data_collection/gazette/spiders/sp_guarulhos.py
@@ -24,7 +24,7 @@ class SpGuarulhosSpider(BaseGazetteSpider):
     def parse(self, response):
         """
         @url http://www.guarulhos.sp.gov.br/diario-oficial/index.php?mes=1&ano=2018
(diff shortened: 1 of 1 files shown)


rennerocha:
For most of the spiders that I saw the contract failing, the solution was to fix just the contract, not the spider code. I don't see that this kind of test is useful and I'd suggest that we stop using contracts this way and maybe thinking in different ways of monitoring the results.

rqclima:
It would be really nice if we can find some other way of monitoring the sanity of these spiders. These tests are really fragile (the spiders are, as the sites can change anytime without notice). And some of the tests look for different data every day, making them even more fragile.

Anyway, it at least tests an important part of the code. And right now it's failing, so I just thought of getting it working again, and then we can think of improving it.

rennerocha:
Contracts on spiders were removed in 3eaeba451ea814661ba19a44e4c6f4412a290bd7 . Discussion related: #44 
As we are not using contracts, this PR is not needed anymore.

PR #129, opened 2019-01-27
Add Spider for RS Gravatai

rodolfolottin:
This PR adds a Spider for the Gravatai city.

#### Questions 

Can someone explain me how can I check if the gazettes that I am crawling are from the legislative or executive legislature? You can check it here: https://gravatai.atende.net/?pg=diariooficial

#### Observartions
I had two problems when working on it. First one was when I tried to do my first commit... The pre-commit hook raised me this error:

```shell
[I] ~/d/diario-oficial ∙ git commit -m 'Update Rs Gravatai docstrings'                                                                                                                                          ᚴ upstream ⚡  rs_gravatai  1M
black....................................................................Failed
hookid: black

Starting diario-oficial_postgres_1 ... done
Starting diario-oficial_redis_1    ... done
Starting diario-oficial_rabbitmq_1 ... done
Usage: black [OPTIONS] [SRC]...
Try "black -h" for help.

Error: Invalid value for "[SRC]...": Path "processing/data_collection/gazette/spiders/rs_gravatai.py" does not exist.
```

Because of that I had to use black locally and used `--no-verify` to write this commit. I think that maybe it can be related to some cached files on my notebook (despite I had run a command that clears all the python cached files).

Second one was when I tried to run my spider:

```shell
019-01-27 01:39:02 [scrapy.core.scraper] ERROR: Error processing {'date': datetime.date(2015, 11, 30),
 'file_urls': ['https://gravatai.atende.net/atende.php?rot=54002&aca=737&processo=download&codigo=162'],
 'files': [{'checksum': '866c92a8386a71ff8a1181ff939eb563',
            'path': 'full/2d6023b5749cac810ed43379ddb499c843f6f0de.php?rot=54002&aca=737&processo=download&codigo=162',
            'url': 'https://gravatai.atende.net/atende.php?rot=54002&aca=737&processo=download&codigo=162'}],
 'is_extra_edition': False,
 'power': 'executive',
 'scraped_at': datetime.datetime(2019, 1, 27, 1, 38, 15, 985416),
 'territory_id': '4309209'}
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/mnt/code/data_collection/gazette/pipelines.py", line 14, in process_item
    item["source_text"] = self.pdf_source_text(item)
  File "/mnt/code/data_collection/gazette/pipelines.py", line 29, in pdf_source_text
    with open(text_path) as file:
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/data/full/2d6023b5749cac810ed43379ddb499c843f6f0de.php?rot=54002&aca=737&processo=download&codigo=162.txt'
```

Seems that the scraper can't find the downloaded file internally on the docker container. Is that ok?

### DIFF
--- a/processing/data_collection/gazette/spiders/rs_gravatai.py
+++ b/processing/data_collection/gazette/spiders/rs_gravatai.py
@@ -18,8 +18,9 @@ class RsGravataiSpider(BaseGazetteSpider):
     def parse(self, response):
         """
         @url https://gravatai.atende.net/?pg=diariooficial
-        @returns requests 68
+        @returns requests 1
         """
+
         last_page_number_css = "#paginacao > ul > li:nth-child(7) > button::attr(value)"
         last_page_number = int(response.css(last_page_number_css).extract_first())
 
@@ -49,8 +50,8 @@ class RsGravataiSpider(BaseGazetteSpider):
 
             code = element.css(".opcoes > button::attr(data-codigo)").extract_first()
             url = (
-                "https://gravatai.atende.net/atende.php?rot=54002&aca=737&"
-                "processo=download&codigo={codigo}".format(codigo=code)
+                "https://gravatai.atende.net/atende.php?rot=54002&aca=737"
+                f"&processo=download&codigo={code}"
             )
 
             yield Gazette(
--- a/processing/data_collection/gazette/spiders/rs_gravatai.py
+++ b/processing/data_collection/gazette/spiders/rs_gravatai.py
@@ -32,7 +32,7 @@ class RsGravataiSpider(BaseGazetteSpider):
     def parse_gazette(self, response):
         """
         @url https://gravatai.atende.net/?pg=diariooficial&pagina=1
-        @returns requests 15
+        @returns items 1
         @scrapes date file_urls is_extra_edition territory_id power scraped_at
         """
 
--- a/processing/data_collection/gazette/spiders/rs_gravatai.py
+++ b/processing/data_collection/gazette/spiders/rs_gravatai.py
@@ -1,0 +1,63 @@
+from dateparser import parse
+from datetime import datetime
+
+from scrapy import Request
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class RsGravataiSpider(BaseGazetteSpider):
+    TERRITORY_ID = "4309209"
+    name = "rs_gravatai"
+    allowed_domains = ["gravatai.atende.net"]
+    start_urls = ["https://gravatai.atende.net/?pg=diariooficial"]
+
+    extra_editions_options = ("Suplementar", "Retificação")
+
+    def parse(self, response):
+        """
+        @url https://gravatai.atende.net/?pg=diariooficial
+        @returns requests 68
+        """
+        last_page_number_css = "#paginacao > ul > li:nth-child(7) > button::attr(value)"
+        last_page_number = int(response.css(last_page_number_css).extract_first())
+
+        for page_number in range(1, last_page_number + 1):
+            yield Request(
+                f"https://gravatai.atende.net/?pg=diariooficial&pagina={page_number}",
+                callback=self.parse_gazette,
+            )
+
+    def parse_gazette(self, response):
+        """
+        @url https://gravatai.atende.net/?pg=diariooficial&pagina=1
+        @returns requests 15
+        @scrapes date file_urls is_extra_edition territory_id power scraped_at
+        """
+
+        for element in response.css(".nova_listagem > .linha"):
+            info = element.css(".info")
+
+            is_extra_edition = (
+                info.css(".tipo::text").extract_first() in self.extra_editions_options
+            )
+
+            date = parse(
+                info.css(".data::text").extract_first(), languages=["pt"]
+            ).date()
+
+            code = element.css(".opcoes > button::attr(data-codigo)").extract_first()
+            url = (
+                "https://gravatai.atende.net/atende.php?rot=54002&aca=737&"
+                "processo=download&codigo={codigo}".format(codigo=code)
+            )
+
+            yield Gazette(
+                date=date,
+                file_urls=[url],
+                is_extra_edition=is_extra_edition,
+                territory_id=self.TERRITORY_ID,
+                power="executive",
+                scraped_at=datetime.utcnow(),
+            )
--- a/CITIES.md
+++ b/CITIES.md
@@ -105,7 +105,7 @@ The municipality id (IBGE code) can be found on [Wikipedia](https://pt.wikipedia
 | 95 | Governador Valadares | :soon: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/19) |
 | 96 | Taboão da Serra | | | |
 | 97 | Santa Maria | | | |
-| 98 | Gravataí | | | |
+| 98 | Gravataí | :soon: | | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/129) |
 | 99 | Várzea Grande | | | |
 | 100 | Abdon Batista | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/141) |
 | 101 | Agrolândia | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/141) |


cuducos:
I think the file name should be `/mnt/data/full/2d6023b5749cac810ed43379ddb499c843f6f0de.pdf` and not `/mnt/data/full/2d6023b5749cac810ed43379ddb499c843f6f0de.php?rot=54002&aca=737&processo=download&codigo=162.txt` but I'm not sure about how to fix that.

Regarding legislative or executive, I'm copying @giovanisleite's reply on Telegram just for documentation:

> Just City Hall -> executive, just Municipal Council  -> legislature. I saw just one gazette from gravatai and looks like executive because all acts come from the mayor or his secretaries.
>
> Normally, I open some gazzetes searching for "vereador", "legislativo", "pref", "sec"

And finally, regarding the Black error, I've never seen this one, but if Black was called manually, no problems ; )

rodolfolottin:
Thanks @cuducos !

So, about black is everything ok.

I'll try to find out what's happening. It seems that it is something related about the files names.

rodolfolottin:
I found the problem!

It seems that the command "pdftotext" that is spawned to run as a subprocess is not escaping special characters. 

Currently a Pipeline is generating the file path like: `/mtn/fulldata/ae782e24ab3dc13dace9853abdf4581d1076b43e.php?rot=54002&aca=737&processo=visualizar&codigo=918`

When trying to convert the pdf to txt the command that is called is:

```shell
pdftotext -layout /mtn/fulldata/ae782e24ab3dc13dace9853abdf4581d1076b43e.php?rot=54002"
```

Instead of:

```shell
pdftotext -layout /mtn/fulldata/ae782e24ab3dc13dace9853abdf4581d1076b43e.php?rot=54002&aca=737&processo=visualizar&codigo=918
```

Note that it is ignoring everything (included) after the first special character (&).

That's why the scrapy can't find the file locally.

One possible solution to resolve this is escaping the pdf path name with:

```python
command = f"pdftotext -layout '{pdf_path}'"
```

What do you think about it? @cuducos , @giovanisleite . Should I open a new PR resolving it?

cuducos:
>  Should I open a new PR resolving it?

Hum… as I said before: I disagree that `ae782e24ab3dc13dace9853abdf4581d1076b43e.php?rot=54002&aca=737&processo=visualizar&codigo=918` should be a proper file name. I think we must treat the URL path to have a file name clean of URL parameters (query string is a known part of the URL scheme, but not meaningful in file systems).

Thus, if by _resolving it_ you mean adding quoates to allow the file name to be something I believe it should not be, I don't agree ; )

rodolfolottin:
I saw that a similar problem occurred with the `Rio de Janeiro` spider and the developer came with this [solution](https://github.com/okfn-brasil/diario-oficial/blob/master/processing/data_collection/gazette/parser.py#L25).

As I see this is exactly what's happening with this spider that I've been working on. On the [`PdfParsingPipeline`](https://github.com/okfn-brasil/diario-oficial/blob/master/processing/data_collection/gazette/pipelines.py#L21) pipeline the item still has the bad formatted url as an attribute.

```python
{'date': datetime.date(2018, 6, 5),
 'file_urls': ['https://gravatai.atende.net/atende.php?rot=54002&aca=737&processo=visualizar&codigo=862'],
 'files': [{'checksum': '0c6b9da582349753ca1f4d5dbd2fcab0',
            'path': 'full/91598d5de3018a741a06a2e11e9b61b866383888.php?rot=54002&aca=737&processo=visualizar&codigo=862',
            'url': 'https://gravatai.atende.net/atende.php?rot=54002&aca=737&processo=visualizar&codigo=862'}],
 'is_extra_edition': False,
 'power': 'executive',
 'scraped_at': datetime.datetime(2019, 1, 31, 23, 41, 13, 745271),
 'territory_id': '4309209'}
```

I checked and the `GazetteFilesPipeline` parser is called and returns the `url` as described on its docstrings. What is still strange is why does the item still has the bad formatted url. Shouldn't the [`GazetteFilesPipeline`](https://github.com/okfn-brasil/diario-oficial/blob/master/processing/data_collection/gazette/parser.py#L36) populate the `item['path']` value?

jvanz:
Hi guys! What's the status of this PR? Do we have news? :)

rodolfolottin:
Hi @jvanz, I have not worked on since that day.

rodolfolottin:
hey @jvanz . I've spent some time debugging the spider to remember why this error is happening.

I've executed the RJ Spider, which [was facing the same error](https://github.com/okfn-brasil/diario-oficial/issues/15), to check for any different behaviour than mine, but I don't know if it's working because I didn't got any pdf downloaded. On that time @brunolellis came [with this solution](https://github.com/brunolellis/diario-oficial/commit/3d643a629afdf8e35c504ceea30b1727278c74f2).

But the error is: the file path still contains query parameters, even tho going through the `GazetteFilesPipeline` parser.

processing/data_collection/gazette/spiders/rs_gravatai.py
```diff
@@ -0,0 +1,63 @@
...
+    name = "rs_gravatai"
+    allowed_domains = ["gravatai.atende.net"]
+    start_urls = ["https://gravatai.atende.net/?pg=diariooficial"]
+
+    extra_editions_options = ("Suplementar", "Retificação")
+
+    def parse(self, response):
+        """
+        @url https://gravatai.atende.net/?pg=diariooficial
+        @returns requests 68
```
jvanz: The number of requests can change. Actually, right now there are 85 pages in the web site. I think you can update the documentation saying it will return a generator of the requests

processing/data_collection/gazette/spiders/rs_gravatai.py
```diff
@@ -0,0 +1,63 @@
...
+
+            date = parse(
+                info.css(".data::text").extract_first(), languages=["pt"]
+            ).date()
+
+            code = element.css(".opcoes > button::attr(data-codigo)").extract_first()
+            url = (
+                "https://gravatai.atende.net/atende.php?rot=54002&aca=737&"
+                "processo=download&codigo={codigo}.pdf".format(codigo=code)
+            )
```
jvanz: There is nothing wrong here, just a suggestion. We use python 3. Thus, you can use the `f-strings`. For example, `f"https://thisismyurl.com/something?codigo={code}"`

jvanz: Have you try returns the original URL, with the `.php`, and adjust the file name in a pipeline item? I'm not 100% sure, but I think we avoid have the `.php` and `pdf` in the `data` folder.

rodolfolottin: I haven't tried that but I think that we can address this problem in a different way. Currently there is a pipeline [responsible for removing the query parameters](https://github.com/okfn-brasil/diario-oficial/blob/master/processing/data_collection/gazette/parser.py) from a url, which is working fine.

This pipeline is also changing the given file extension from `.pdf` to `.php` when [it overrides the media_ext variable](https://github.com/okfn-brasil/diario-oficial/blob/master/processing/data_collection/gazette/parser.py#L33) based on the path resulted from the parsing.

```python
ipdb> url
'https://gravatai.atende.net/atende.php?rot=54002&aca=737&processo=download&codigo=1053.pdf'
ipdb> urlparse(url)
ParseResult(scheme='https', netloc='gravatai.atende.net', path='/atende.php', params='', query='rot=54002&aca=737&processo=download&codigo=1053.pdf', fragment='')
```

Do you know if we have any other spiders taking advantage of this logic? I've executed the Rio de Janeiro spider but it didn't downloaded any pdf files. Because if not, I think that we may not need this if statement content any more. Since, for this specific spider, which is downloading pdf files with query parameters in its url, the media_ext variable already comes without the query parameters. What do you think?

jvanz: I have an idea. Instead of remove the pipeline item we can improve it to be more generic. I have took a glance in the scrapy source code and  noticed that the response object has the [headers](https://github.com/scrapy/scrapy/blob/a6bf895474ade06efc818582db55beb7a2545e32/scrapy/http/response/__init__.py#L21). So, we can ignore file  extensions because it does not mean to much, specially on Linux. We can use the `Content-Type` to define the file extension. If the content type is "not supported", we can use the file extension and print a warning. What do you think? :)

Furthermore, I still think you should remove that .pfd from the URL. This data is storage in the database and it can be really hard to somebody else to remember that the .pdf part is not from the original URL. Thus, we keep a better historical info. By the way, it's kind of surprise to me a server responding for a modified URL.

jvanz: Take a look on this lib as well:
```
>>> import mimetypes as mt
>>> mt.guess_all_extensions("application/pdf")
['.pdf']
```

jvanz: One more info that I found. I've already sent this to you via Telegram. But I'll put here as well for documentation. To validate my previous idea, I wrote a simple `print` statement to see the response headers:

```
{b'Date': [b'Tue, 28 Jan 2020 22:16:39 GMT'], b'Server': [b'Apache'], b'Expires': [b'Mon, 26 Jul 1997 05:00:00 GMT'], b'Cache-Control': [b'must-revalidate, post-check=0, pre-check=0'], b'Pragma': [b'public'], b'Data-Servidor': [b'1580249799000'], b'Content-Disposition': [b'attachment; filename="Edi\xe7\xe3oOrdin\xe1ria666.pdf"'], b'Strict-Transport-Security': [b'max-age=31536000; includeSubDomains; preload'], b'Content-Type': [b'application/octet-stream']} 
```

As you can see the `Content-Type` is not very helpful in this case. But it can be for other spiders. But for now, let's focus on yours. So, even if `Content-Type` does not have anything useful, because `application/octet-stream` can be anything, take a look in the `Content-Disposition`. There we can see the `filename`. Maybe is worth full try to use it to get the file type.  As a fallback we can  continue use the file extension as it is now.

rodolfolottin: I was struggling a bit because scrapy was caching the first requests and the response was always empty, but I did manage to debug the response object content. As soon as I have some free time I'll work on it! Thanks for your support @jvanz

jvanz:
I've had an idea during my lunch. Instead of trying to get the file extension, let's ignore it. In the end of the day, it does not make to much difference on Linux. Take a look in a simple example:

```
jvanz@earth:~> vim mytest.doc
jvanz@earth:~> file mytest.doc 
mytest.doc: ASCII text
```

My suggestion is keep the whatever name the `file_path` returns. Then, change the step which extract the text to use the `python-magic` lib to find the file type. For example:

```
>>> import magic
>>> magic.from_file("mytest.doc", mime=True)
'text/plain'
```

Other option is to add  one more step to do this check.

rodolfolottin:
> I've had an idea during my lunch. Instead of trying to get the file extension, let's ignore it. In the end of the day, it does not make to much difference on Linux. Take a look in a simple example:
> 
> ```
> jvanz@earth:~> vim mytest.doc
> jvanz@earth:~> file mytest.doc 
> mytest.doc: ASCII text
> ```
> 
> My suggestion is keep the whatever name the `file_path` returns. Then, change the step which extract the text to use the `python-magic` lib to find the file type. For example:
> 
> ```
> >>> import magic
> >>> magic.from_file("mytest.doc", mime=True)
> 'text/plain'
> ```
> 
> Other option is to add one more step to do this check.

hei @jvanz , I've made the changes that we've talked about. You can check it here: #156 . Soon it gets reviewed and accepted we can then move on with this spider.

rodolfolottin:
Hei @jvanz !

I've rebased the branch, ran the spider and everything is working fine.

Sorry for the delay and thanks for all your help!

PR #130, opened 2019-03-08
WIP: Recife/PE spider

andersonberg:
I'm trying to build a spider to get the documents from Recife.

However, the website has some kind of mecanism to generate the pdf before the user can download it. So the spider gets `500 Internal Server Error` every time it tries to download the pdf.

`
2019-03-08 16:57:48 [scrapy.downloadermiddlewares.retry] DEBUG: Gave up retrying <GET http://200.238.101.22/docreader/saveasfile.aspx?cache=cache/56881009833354&arq=Arquivo.pdf> (failed 3 times): 500 Internal Server Error
`

I'm not sure how to fix this. I would appreciate some help here.

### DIFF
--- a/CITIES.md
+++ b/CITIES.md
@@ -16,7 +16,7 @@ The municipality id (IBGE code) can be found on [Wikipedia](https://pt.wikipedia
 | 6 | Belo Horizonte | :soon: | | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/33) |
 | 7 | Manaus | :white_check_mark: | | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/51) |
 | 8 | Curitiba | :soon: | | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/42) |
-| 9 | Recife | :soon: | | | |
+| 9 | Recife | :soon: | | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/130) |
 | 10 | Porto Alegre | :white_check_mark: | :white_check_mark: | | |
 | 11 | Goiânia | :white_check_mark: | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/6) |
 | 12 | Belém | | | | |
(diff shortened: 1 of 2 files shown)


fbormann:
@andersonberg  I will try to help you out, can we get in touch? My e-mail is felipe.bormann@gmail.com

jvanz:
Hi guys! What's the status of this PR?

andersonberg:
I'm sorry, I didn't get in touch with Felipe. My fault.

jvanz:
> I'm sorry, I didn't get in touch with Felipe. My fault.

No reason to apologize. I'm just walking throu some PRs and asking if there is someone working on them. Furthermore, I believe it would be better if you discuss the change in the PR. Thus, future contributor can review it and learn. ;)

giuliocc:
@andersonberg 

This website uses a temporary url for the documents. To request the temporary url to be made this way you would need to make two intermediary requests:
 
1. A GET to fetch the exporting menu;
2. A POST requesting the document to be exported to PDF (the previous request is need because the website is made with ASP.NET and checks `_VIEWSTATE` and stuff like that.

After these, your document should be available at the url you were trying.

... but there's a catch :)

As you've probably noticed. Exporting these files to PDF takes a looong time. Sometimes it's so long that the server never bothers to send you a response for the second intermediary request mentioned above. I've tried making the `DOWNLOAD_TIMEOUT` as long as 20 minutes to no avail. The server simply ignores you. Some light shined upon my crying self for some time when I discovered that if you are able to make the request even if you don't get a response, the url is available and you can download it. But then you have to decide when is the right moment to do that. And as the time to export varies so much depending on the size of the gazette, I've crawled backed into the darkness.

Although, you can still make the spider work. If you just ignore the exporting part hehehehe. Not as it could be with this implementation , though. And it brings up more challenges do be solved. I've implemented it in #170. Check it out, maybe give me some tips.

Hope this helped!

jvanz:
As the PR https://github.com/okfn-brasil/diario-oficial/pull/170 has been merged. I think we can close this one.

PR #135, opened 2019-11-12
Fecam

jvanz:
This is the initial version of a spider to get the data from [FECAM](https://www.fecam.org.br/) website. I've created a spider class which able to understand the fecam website and navigate thou the pages and extract the gazette file. 

Most of the file are docs. Thus, I've also added a new step in the pipeline to convert the doc into pdf. For that, i used Libreoffice writer. Its the best option in the top of my mind, if you have a better idea, let me know. 

I'll continue testing and adding the missing cities which have their gazette in the website. Currently, I'm just testing with my city

The files are being downloaded from: https://diariomunicipal.sc.gov.br/site/

### DIFF
--- a/processing/data_collection/gazette/pipelines.py
+++ b/processing/data_collection/gazette/pipelines.py
@@ -54,7 +54,9 @@ class ExtractTextPipeline:
         elif self.is_pdf(item["files"][0]["path"]):
             item["source_text"] = self.pdf_source_text(item)
         else:
-            raise "Unsupported file type"
+            raise Exception(
+                "Unsupported file type: " + self.get_extension(item["files"][0]["path"])
+            )
 
         for key, value in item["files"][0].items():
             item[f"file_{key}"] = value
@@ -101,3 +103,10 @@ class ExtractTextPipeline:
         """
         filepath = filepath.lower()
         return filepath.endswith("doc") or filepath.endswith("docx")
+
+    @staticmethod
+    def get_extension(filename):
+        """
+        Returns the file's extension
+        """
+        return filename[filename.rfind(".") :]
--- a/processing/data_collection/gazette/pipelines.py
+++ b/processing/data_collection/gazette/pipelines.py
@@ -67,9 +67,9 @@ class ExtractTextPipeline:
         Gets the text from pdf files
         """
         pdf_path = os.path.join(FILES_STORE, item["files"][0]["path"])
-        command = f"pdftotext -layout {pdf_path}"
-        subprocess.run(command, shell=True, check=True)
         text_path = pdf_path + ".txt"
+        command = f"pdftotext -layout {pdf_path} {text_path}"
+        subprocess.run(command, shell=True, check=True)
         with open(text_path) as file:
             return file.read()
 
--- a/processing/data_collection/gazette/pipelines.py
+++ b/processing/data_collection/gazette/pipelines.py
@@ -78,13 +78,7 @@ class DocToPdfPipeline:
         # use libreoffice writer to convert
         command = f"lowriter --convert-to pdf --outdir {FILES_STORE}/full {doc_path}"
         subprocess.run(command, shell=True, check=True)
-        if doc_path.endswith("doc"):
-            pdf_path = doc_path[:-3] + "pdf"
-        elif doc_path.endswith("docx"):
-            pdf_path = doc_path[:-4] + "pdf"
-        else:
-            pdf_path = doc_path + ".pdf"
-        os.unlink(doc_path)
+        pdf_path = doc_path + ".pdf"
         # update to the new file path and its checksum
         item["files"][0]["path"] = pdf_path
         item["files"][0]["checksum"] = self.calculate_md5sum(pdf_path)
--- a/processing/data_collection/gazette/pipelines.py
+++ b/processing/data_collection/gazette/pipelines.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+import hashlib
 
 from database.models import Gazette, initialize_database
 from scrapy.exceptions import DropItem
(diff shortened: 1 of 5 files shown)


vitorbaptista:
Hello @jvanz! Nice PR. Just so I understand, why are you converting a DOC to a PDF? Is it just so we can reuse the PDF parsing pipeline?

I'm not from the core team, but it is a strange strategy to convert from a "more machine-readable" format (doc) to a less readable format (pdf) before parsing the data. IMHO, it would be better to change the pipeline to support not only PDFs, but also DOCs. Especially because I imagine there will be other formats in the future (e.g. TXT?), and this current requirement that the gazettes are PDFs will have to be changed anyway.

I imagine the `PdfParsingPipeline` becoming something like a `TextExtractorPipeline`, and supporting PDFs, DOC, etc. Have you tried something like this?

jvanz:
> Hello @jvanz! Nice PR. Just so I understand, why are you converting a DOC to a PDF? Is it just so we can reuse the PDF parsing pipeline?

Yes
> 
> I'm not from the core team, but it is a strange strategy to convert from a "more machine-readable" format (doc) to a less readable format (pdf) before parsing the data. IMHO, it would be better to change the pipeline to support not only PDFs, but also DOCs. Especially because I imagine there will be other formats in the future (e.g. TXT?), and this current requirement that the gazettes are PDFs will have to be changed anyway.
> 
> I imagine the `PdfParsingPipeline` becoming something like a `TextExtractorPipeline`, and supporting PDFs, DOC, etc. Have you tried something like this?

I decided to not extract the txt from the doc because I didn't find a good lib for that. I know there is the python-docx library. But from previous experiences, where it was not able to read all docish files, I decided to convert to pdf. However, if you think that the way to go, I can update the PR. ;)

vitorbaptista:
@jvanz Regarding extracting text from DOC, have you tried [Apache Tika](https://tika.apache.org/)? I've never used it, but heard good things. If it works well, I think it's the ideal tool, as it supports dozens of file formats (including PDFs). What do you think, @cuducos / @Irio ?

jvanz:
> @jvanz Regarding extracting text from DOC, have you tried [Apache Tika](https://tika.apache.org/)? I've never used it, but heard good things. If it works well, I think it's the ideal tool, as it supports dozens of file formats (including PDFs). What do you think, @cuducos / @Irio ?

I never have tried Apache Tika. Actually, I never have hear about it. It seems interesting indeed

processing/data_collection/gazette/spiders/base.py
```diff
@@ -1,6 +1,10 @@
 # -*- coding: utf-8 -*-
 import dateparser
 import scrapy
+import re
+from datetime import datetime
+
+from gazette.items import Gazette
```
cuducos: These blocks are not following [PEP8 recommendations for imports](https://www.python.org/dev/peps/pep-0008/#imports), that is to say: one block for native modules, another one for third-party libraries and the third one for local modules:

```python
import re
from datetime import datetime

import dateparser
import scrapy

```

processing/data_collection/gazette/pipelines.py
```diff
@@ -61,3 +62,51 @@ def process_item(self, item, spider):
...
+        item["files"][0]["checksum"] = self.calculate_md5sum(pdf_path)
+        return item
+
+    @staticmethod
+    def is_doc(filepath):
+        """
+        If the file path ends with doc or docx returns True. Otherwise,
+        returns False
+        """
+        return filepath.endswith("doc") or filepath.endswith("docx")
```
cuducos: Maybe we should use `.lower()` to normalize `filepath`? This might fail with `DOCUMENT.DOCX` for example.

processing/data_collection/gazette/pipelines.py
```diff
@@ -61,3 +62,51 @@ def process_item(self, item, spider):
...
+        doc_path = os.path.join(FILES_STORE, item["files"][0]["path"])
+        # use libreoffice writer to convert
+        command = f"lowriter --convert-to pdf --outdir {FILES_STORE}/full {doc_path}"
+        subprocess.run(command, shell=True, check=True)
+        if doc_path.endswith("doc"):
+            pdf_path = doc_path[:-3] + "pdf"
+        elif doc_path.endswith("docx"):
+            pdf_path = doc_path[:-4] + "pdf"
+        else:
+            pdf_path = doc_path + ".pdf"
```
cuducos: I think we should just add the `.pdf` independent of the original name/extension (so we can track if the file was `.doc` or a `.docx` if needed).

processing/data_collection/gazette/pipelines.py
```diff
@@ -61,3 +62,51 @@ def process_item(self, item, spider):
...
+        os.unlink(doc_path)
```
cuducos: Please, don't. We need to keep the original files for accountability. Otherwise, people might argue we _edited_ the file when converting to PDF. The only way to be accountable is to keep the original file we've got from the original source.

cuducos:
Many thanks for the contribution, @jvanz! Really great code and solutions IMHO.

I don't know Apache Tika either, but I have the feeling that it would be more trustworthy to extract text from Word document than from PDF – I mean, the quality of the extracted text would be considerably better coming from `.doc` than from `.pdf`. But it's only a guess.

In other ways, I would consider extracting the text from the Word document and not from the converted PDF.

jvanz:
I believe that all issues are solved in the last commits. Related to the Apache Tika, do you agree to leave this for a separate PR? I can work on that before are adding other spiders to other cities.

processing/data_collection/gazette/pipelines.py
```diff
@@ -64,44 +43,64 @@ def process_item(self, item, spider):
...
+        """
+        Gets the text from pdf files
+        """
+        pdf_path = os.path.join(FILES_STORE, item["files"][0]["path"])
+        command = f"pdftotext -layout {pdf_path}"
+        subprocess.run(command, shell=True, check=True)
+        if ".pdf" in pdf_path:
+            text_path = pdf_path.replace(".pdf", ".txt")
+        else:
+            text_path = pdf_path + ".txt"
```
vitorbaptista: Similarly to how you did in `doc_source_text()`, I think here would be better to just append `.txt` to the filename, getting something like `foobar.pdf.txt`, as @cuducos suggested this would keep the information that this file was originally a PDF.

vitorbaptista:
Apart from the small comment I added inline, this is looking great! I haven't tried running it, but it seems ready to go :+1:

processing/data_collection/gazette/pipelines.py
jvanz: Done! ;)

jvanz:
So guys, what do you think? Is the PR in the right path? :)

processing/data_collection/gazette/spiders/base.py
```diff
@@ -11,3 +16,75 @@ def __init__(self, start_date=None, *args, **kwargs):
...
+        Get the last page number available in the pages navigation menu
+        """
+        href = response.xpath(
+            "/html/body/div[1]/div[3]/div[4]/div/div/ul/li[14]/a/@href"
+        ).get()
+        result = re.search("Search_page=(\d+)", href)
+        if result is not None:
+            return int(result.groups()[0])
+
+    def get_gazzete(self, document):
```
vitorbaptista: Typo: gazzete -> gazette

PR #139, opened 2019-11-21
Fix pagination order - sc_florianopolis

feliperuhland:
- Fix spider pagination order - ac7e484
- Ignore gazettes before the start_date variable - 4246a1e

### DIFF
--- a/processing/data_collection/gazette/spiders/sc_florianopolis.py
+++ b/processing/data_collection/gazette/spiders/sc_florianopolis.py
@@ -21,7 +21,7 @@ class ScFlorianopolisSpider(BaseGazetteSpider):
             year, month = str(target.year), str(target.month)
             data = dict(ano=year, mes=month, passo="1", enviar="")
             yield FormRequest(url=self.URL, formdata=data, callback=self.parse)
-            target = target + relativedelta(months=1)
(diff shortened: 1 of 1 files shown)


sergiomario:
@giovanisleite Can you review this?

processing/data_collection/gazette/spiders/sc_florianopolis.py
```diff
@@ -21,16 +21,23 @@ def start_requests(self):
             year, month = str(target.year), str(target.month)
             data = dict(ano=year, mes=month, passo="1", enviar="")
             yield FormRequest(url=self.URL, formdata=data, callback=self.parse)
-            target = target + relativedelta(months=1)
+            if target <= self.start_date:
```
FelipeFin: I think it would be safer to check for **self.start_date** before trying to use it, as this attribute could be None and it would break the spider.

jvanz:
Besides the @FelipeFin's comment, LGTM

processing/data_collection/gazette/spiders/sc_florianopolis.py
feliperuhland: Nice catch!

PR #141, opened 2019-11-29
Santa Catarina cities

jvanz:
As we have a base spider for the FECAM web site, now we can all the cities which have their gazettes there. Brace ourselves! This PR has a lot of cities. 

I know that are a lot of commit and files. But, they are very similar because all of them are based in the same class. If you think that I should split the PR, let me know. 

https://github.com/okfn-brasil/diario-oficial/issues/134

Thanks!


jvanz:
> Awesome job!
> Perhaps it would be nice to add all those cities to the CITIES.md as well?

Yeap. I will. :)

jvanz:
> Awesome job!
> Perhaps it would be nice to add all those cities to the CITIES.md as well?

Done! ;)

PR #142, opened 2019-11-30
base.py: makes FECAM spider run faster

jvanz:
Updates the FECAM spider to make it run faster. For that, instead of
return the next page after get all the gazettes files, the first request
find all the pages and returns a request object for each one. Thus,
scrapy can parallelize as much as possible.

### DIFF
--- a/processing/data_collection/gazette/spiders/base.py
+++ b/processing/data_collection/gazette/spiders/base.py
@@ -24,24 +24,30 @@ class FecamGazetteSpider(scrapy.Spider):
     total_pages = None
 
     def start_requests(self):
-        if self.total_pages is None:
-            yield scrapy.Request(
-                f"{self.URL}?q={self.FECAM_QUERY}", callback=self.parse
+        yield scrapy.Request(
+            f"{self.URL}?q={self.FECAM_QUERY}", callback=self.first_parse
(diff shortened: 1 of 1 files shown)


processing/data_collection/gazette/spiders/base.py
```diff
@@ -24,24 +24,30 @@ class FecamGazetteSpider(scrapy.Spider):
...
 
     def start_requests(self):
-        if self.total_pages is None:
-            yield scrapy.Request(
-                f"{self.URL}?q={self.FECAM_QUERY}", callback=self.parse
+        yield scrapy.Request(
+            f"{self.URL}?q={self.FECAM_QUERY}", callback=self.first_parse
+        )
+
+    def first_parse(self, response):
```
vitorbaptista: Maybe something like `parse_pages` (or `parse_pagination`) and `parse_gazettes` would be a better name? `first_parse` doesn't tell me much.

vitorbaptista:
Apart from my comment on naming, this looks great :+1:

processing/data_collection/gazette/spiders/base.py
jvanz: Fixed!

PR #143, opened 2019-12-01
pipelines.py: support txt file

jvanz:
Some gazettes have text files. This PR updates the ExtractTextPipeline pipeline step to extract data from text files.


jvanz:
> I don't like the hardcoded ISO-8859-1 encoding, as we'll probably face a UTF-8 text sometime. I think the only way to avoid these encoding errors is to standardize all files into a single encoding (probably UTF-8) when we save them into our system. So the responsibility of converting the encoding lies on each crawler.
> 
> However, I imagine ISO-8859-1 is more common than UTF-8 in Brazil. Also, this encoding problem haven't happened yet, so we can fix it when it appears. It's just something to keep in mind.
> 
> Great work! +1

I agree. I was thinking about that too, but I did not find any solution good enough. So, I leave  the issue for the future. ;P

PR #149, opened 2019-12-12
Update Apache Tika

jvanz:
Updates the Apache Tika version used. The old version is not available
anymore

Issue https://github.com/okfn-brasil/diario-oficial/issues/147

Signed-off-by: José Guilherme Vanz <jvanz@jvanz.com>

PR #150, opened 2019-12-12
Makefile: helper commands

jvanz:
Adds some helpers functions in the Makefile to open the psql, run spiders
and delete the gazette files


feliperuhland:
LGTM, but the build is broken. I think #149 will fix it.

jvanz:
> LGTM, but the build is broken. I think #149 will fix it.

Yeap, that's right

PR #155, opened 2020-01-15
WIP: Processing pipeline

jvanz:
This is a **POC** which I'm working on. The idea is to create a processing pipeline to extract and store useful info found in the gazette files. 

In the current version, the gazette text are processed to find all the CNPJ present in the text. Once we have that, a graph database is updated to show that relationship between a city and a CNPJ. Furthermore, the [brasil.io](https://brasil.io/home) API is used to get the name for the CNPJs. The following image is a example of the graph:

Screenshot from 2020-01-14 22-14-32

On this image, all the blue nodes are cities and the green ones are CNPJs. It's also possible to see the CNPJ info from a node in the lower right corner.

There are a lot of room for improvement, but I'll leave such enhancements to the future. First, I would like to see what the community thinks about the idea. But I already have some ideas to improve the pipeline:

- Use cache to avoid query brasil.io and graph database all the time
- Try reduce the amount to request to the graph database doing more in each transaction
- Improve the python code as a whole. 
- Improve the messages between service to avoid to many reads from the Kafka topic.
- Validate what is the best graph database for our use case (Neo4J, Janus, DGraph, etc). The current version uses the DGraph. I choose it just because I was curious about it. hehe
- Considering the previous item. Decouple the code to interact with the graph database
- Continue using Kafka ?
- Among other that we can find. :)
- Add the owner/partners of the CNPJs in the database. https://brasil.io/dataset/socios-brasil/socios
- Unit tests

Feedback is welcome!


jvanz:
I've committed the `municipios.csv` just for a matter of convinience to everybody to test. But I think we can add a script (or a Makefile target) to download it automatically with no need to commit it in the repo. ;)

jvanz:
In the https://github.com/okfn-brasil/diario-oficial/pull/155/commits/95596a40aa4f49c0c00d8646028a16e2fb15ca86  I've added the companies business partners in the graph database.

PR #156, opened 2020-02-08
Update scrapy library version

rodolfolottin:
This PR updates the scrapy library version and changes how the Extractor pipeline identifies, for a given file path, its file type.

#### Why? 

A few years ago one contributor identified on an issue (#15) a problem when a download url contained query params. To resolve it [he created a temporary pipeline](https://github.com/brunolellis/diario-oficial/commit/75b8b7ce7512d3e08ebba8af0ac127d59a216aa6) in which he overwrote the scrapy file_path method to escape the string query parameters. Besides it was working properly for his use case (RJ - Rio de Janeiro spider), on [a spider that I've been trying to complete it wasn't.](https://github.com/okfn-brasil/diario-oficial/pull/129#discussion_r371547950).

This year me and @jvanz started to talk about workarounds to solve this problem and we've found that by updating the scrapy library we would not longer face it. Also, this PR changes how the file type is identified, by using a standard command from the Unix operating systems interfaced by [python-magic](https://pypi.org/project/python-magic/) library.

So, when this pull request gets accepted we will not longer have any problems related to download urls with query parameters. #129


processing/data_collection/gazette/pipelines.py
```diff
@@ -93,40 +94,50 @@ def doc_source_text(self, item):
...
         """
-        If the file path ends with doc,  docx or odt returns True. Otherwise,
+        If the file type is doc or similar returns True. Otherwise,
         returns False
         """
-        extension = self.get_extension(filepath)
-        return extension == "doc" or extension == "docx" or extension == "odt"
+        file_types = [
+            "application/msword",
+            "application/octet-stream",
```
jvanz: I think you can remove the `application/octet-stream`. It's to generic. You can also add `application/vnd.openxmlformats-officedocument.wordprocessingml.document`, it's the docx mime type. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types

jvanz:
@cuducos what do you think?

processing/data_collection/gazette/pipelines.py
rodolfolottin: Gotcha! 

I don't remember for which file type I've added `application/octet-stream`. Maybe there was something wrong with the file sample that I downloaded to test it.

Thanks for the review!

PR #162, opened 2020-03-06
scrapy-deltafetch: avoid crawl the same URL

jvanz:
Adds the scrapy-deltafetch middleware. It allow us to drop item which
already has been crawled. It uses the URL to identify if an item has
already been download. Even if it can causes issues if the same URL
points to a different data in the future, it sounds reasonable to add it
to avoid waste of resources. Furthermore, it is always possible to set
the middleware to reset and allow all the item to be crawled again.

https://github.com/okfn-brasil/diario-oficial/issues/86

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ ENV/
 
 # dolphin
 .directory
+
+# scrapy deltafetch middleware directory
+data/deltafetch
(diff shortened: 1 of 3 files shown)


jvanz:
@rennerocha FYI

PR #166, opened 2020-04-08
Update Apache Tika to 1.24 in Dockerfile

AndreAngelucci:
The [link](http://ftp.unicamp.br/pub/apache/tika/tika-app-1.23.jar) to download Apache Tika 1.23 is broken on Dockerfile. Can we update to 1.24?


jvanz:
Thanks for that! I've already faced the same issue. I've tried to find a stable link, but no success. =/

jvanz:
Oh! It's your first contribution. Thanks again! 

external-content duckduckgo com

PR #168, opened 2020-04-29
Created script to handle Apache Tika download from Unicamp FTP

### LINKED ISSUE #167: Fix Apache Tika URL
Currently, the URL to download the Apache Tika in our container image to run the spider changes with some annoying frequency. Every time a version is released the old one is deleted and our URL is not valid anymore. We need to find a way to make it more stable. In the current version, the URL is maintained by Unicamp. 

On top of my head I think we can look for a URL which keeps more than one version at time. Thus, the binary will be available for more time.

jaswdr:
This PR fix #167 by creating a script to handle the problem with filename changing in remote Unicamp FTP

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 setup:
 	cp .env.example .env
 	docker-compose pull
-	docker-compose build
+	docker-compose build --build-arg dns=1.1.1.1
 	make seed
 	@if [ -z $$VIRTUAL_ENV ]; then \
 		pip install --user pre-commit; \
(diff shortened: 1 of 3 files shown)


jaswdr:
ping @jvanz

jvanz:
Thanks @jaswdr  for your first contribution!

giphy
 
 But I think we can go with other solution from #169 , it's more simple and easy to maintain.

jaswdr:
closing in favor of #169

PR #171, opened 2020-05-17
Fix LOCAL_USER_ID envvar behaviour

giuliocc:
The `env_file` option in the `docker-compose.yml` exports the envvars when running the container but not when building it.

Unfortunately, the Compose's documentation does not cover it well (or I couldn't find it there). The best I could find to link here was this note

image

on the [environment section](https://docs.docker.com/compose/compose-file/#environment) that tells us to use the `args` sub-option under the `build` option to define envvars.

To complement this, if no Compose is used, the `--build-arg` option of `docker build` has to be used to export envvars at build time. As specified in the [`ARG` statement documentation](https://docs.docker.com/engine/reference/builder/#arg).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,10 @@ services:
     image: postgres:11.3-alpine
 
   processing:
-    build: processing
+    build:
+      context: processing
+      args:
+        - LOCAL_USER_ID
     depends_on:
       - postgres
     env_file:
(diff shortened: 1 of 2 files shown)


jvanz:
I'm curious, did you face some issue which make you did this change? What's the root cause of the change?

giuliocc:
Yep, I had id 1001 assigned to my user and couldn't get the permissions right through docker-compose :/

jvanz:
The changes seems to be fine. Can you please add some documentation in the README about your problem and how to use ? It's not complex, but I think is nice to have at least couple of lines  about that to help the users.

PR #60, opened 2018-05-29
Add spider for Natal/RN

dannnylo:
In this PR we add the spider to collect data from [Natal/RN gazette](http://www.natal.rn.gov.br/dom/)

PR #192, opened 2020-06-26
Add spider for Natal/RN

### LINKED ISSUE #189: Natal spider
Add spider for Natal

victor-torres:
This pull request takes over development started by @dannnylo on #60 
This should fix #189

### DIFF
--- a/processing/data_collection/gazette/spiders/rn_natal.py
+++ b/processing/data_collection/gazette/spiders/rn_natal.py
@@ -1,0 +1,46 @@
+import re
+import dateparser
+import w3lib.url
+
+from datetime import datetime
+from scrapy import Request, Spider, FormRequest
+from gazette.items import Gazette
+
+
+class RnNatalSpider(Spider):
+    GAZETTE_ELEMENT_CSS = '#texto a'
+    MUNICIPALITY_ID = '2408102'
+
+    allowed_domains = ['www.natal.rn.gov.br']
+    name = 'rn_natal'
+
+    def start_requests(self):
+        base_url = 'http://www.natal.rn.gov.br/dom/'
+
+        for year in range(2015, datetime.now().year + 1):
+            for month in range(1, 13):
+                print([year, month])
+                data = dict(ano=str(year), mes=str(month), list='Listar')
+                yield FormRequest(url=base_url, formdata=data, callback=self.parse)
+
+    def parse(self, response):
+        """
+        @url http://www.natal.rn.gov.br/dom/
+        @returns items 1
+        @scrapes date file_urls is_extra_edition municipality_id power scraped_at
+        """
+
+        for element in response.css(self.GAZETTE_ELEMENT_CSS):
+            url = response.urljoin(element.css('::attr(href)').extract_first())
+            link_text = element.css('::text').extract_first()
+            date = dateparser.parse(link_text.split(' - ')[-1], languages=['pt']).date()
+            extra_edition = ("Extra" in link_text)
+
+            yield Gazette(
+                date=date,
+                file_urls=[url],
+                is_extra_edition=extra_edition,
+                municipality_id=self.MUNICIPALITY_ID,
+                power='executive',
+                scraped_at=datetime.utcnow()
+            )
--- a/processing/data_collection/gazette/spiders/rn_natal.py
+++ b/processing/data_collection/gazette/spiders/rn_natal.py
@@ -31,6 +31,6 @@ class RnNatalSpider(BaseGazetteSpider):
                 file_urls=[file_url],
                 is_extra_edition=extra_edition,
                 territory_id="2408102",
-                power="executive",
+                power="executive_legislative",
                 scraped_at=datetime.utcnow(),
             )
--- a/processing/data_collection/gazette/spiders/rn_natal.py
+++ b/processing/data_collection/gazette/spiders/rn_natal.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 import dateparser
-from scrapy import Request, FormRequest
+from scrapy import FormRequest
 
 from gazette.items import Gazette
 from gazette.spiders.base import BaseGazetteSpider
--- a/processing/data_collection/gazette/spiders/rn_natal.py
+++ b/processing/data_collection/gazette/spiders/rn_natal.py
@@ -23,7 +23,7 @@ class RnNatalSpider(BaseGazetteSpider):
         for entry in response.css('#texto a'):
             file_url = response.urljoin(entry.css('::attr(href)').get())
             title = entry.css('::text').get()
-            date = dateparser.parse(title.split(' - ')[-1], languages=['pt']).date()
+            date = dateparser.parse(title.split('-')[-1], languages=['pt']).date()
             extra_edition = 'Extra' in title
 
             yield Gazette(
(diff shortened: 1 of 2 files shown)


jvanz:
@giuliocc how about review this PR ? ;)

giuliocc:
Sure. Thanks :)

processing/data_collection/gazette/spiders/rn_natal.py
```diff
@@ -0,0 +1,36 @@
+from datetime import datetime
+
+import dateparser
+from scrapy import Request, FormRequest
```
giuliocc, line 4:
```suggestion
from scrapy import FormRequest
```
This can be removed since it's not being used.

processing/data_collection/gazette/spiders/rn_natal.py
```diff
@@ -0,0 +1,36 @@
...
+
+class RnNatalSpider(BaseGazetteSpider):
+
+    name = "rn_natal"
+    allowed_domains = ["www.natal.rn.gov.br"]
+
+    def start_requests(self):
+        base_url = "http://www.natal.rn.gov.br/dom/"
+        for year in range(2003, datetime.now().year + 1):
+            for month in range(1, 13):
```
giuliocc, line 18: If you use `dateutil.rrule` you can express your intent much more and have more power on the upper and lower date limits. [Here is the examples page](https://dateutil.readthedocs.io/en/stable/examples.html#rrule-examples), look for `MONTHLY`.

processing/data_collection/gazette/spiders/rn_natal.py
```diff
@@ -0,0 +1,36 @@
...
+            file_url = response.urljoin(entry.css("::attr(href)").get())
+            title = entry.css("::text").get()
+            date = dateparser.parse(title.split("-")[-1], languages=["pt"]).date()
+            extra_edition = "Extra" in title
+
+            yield Gazette(
+                date=date,
+                file_urls=[file_url],
+                is_extra_edition=extra_edition,
+                territory_id="2408102",
```
giuliocc, line 33: I think you don't lose anything putting the `TERRITORY_ID` attribute on the spider, it makes it easier on the eye and makes it easier for refactoring and scripting. What do you think about it?

processing/data_collection/gazette/spiders/rn_natal.py
```diff
@@ -0,0 +1,36 @@
...
+                power="executive",
```
giuliocc, line 34:
```suggestion
                power="executive_legislative",
```
I randomed a month and found that in 05/04/2006 and 18/04/2006 there are sections dedicated to the Municipal Chamber. This indicates that this gazette is related to both the Executive and the Legislative power. I couldn't find a way to differentiate whether a gazette includes only Executive information or both. Thus, I am suggesting the change of this field to be always `"executive_legislative"`.

processing/data_collection/gazette/spiders/rn_natal.py
victor-torres: I don't think this territory code should change. Therefore, I don't see any gain in moving it to a constant. If you really want me to do it, I can move it.

processing/data_collection/gazette/spiders/rn_natal.py
victor-torres: I appreciate the link. This is a cool function, although I think `range` plays fine here. It's not something very complex.

victor-torres: Let me know if you believe it should be changed anyway.

victor-torres:
Thank you for your review, @giuliocc.

I've tried not to make substantial changes to the original author's source code.

Also, this is a very simple spider. I've sketched a longer and more organized (read complex) version of this spider but I don't think it's worth doing it for a spider that's so small and which complexity is very low. 

In this case, it doesn't pay the effort to write and maintain a more complex spider in the long run. Especially because when spiders change, they probably completely break the old extraction logic. If that happens with small spiders like this, it's often quicker to write a new one from scratch than trying to fix the old one :)

Regarding documentation, well, there might be a knowledge bias playing its role here, but I think the spider is pretty self-documented. It has ~30 lines and it's very easy to see what's going on here. Adding docs would be a little bit redundant in this case, but again, that's my opinion and it might be biased.

I could have dedicated more attention to the commit messages, but feel free to squash it before the merge if it gets approved.

Wish I had more time to dedicate to this project :)

Cheers,
Victor Torres

processing/data_collection/gazette/spiders/rn_natal.py
giuliocc: I understand and agree that it shouldn't change much, at least for the capitals we can have a good amount of confidence that it will not change.

I think that making it a spider attribute also makes it easier for accessing it in a pipeline or something of sort (like allowing this field to be filled automatically if it's missing). This argument is not following YAGNI but I am trying to provide some reasons for "why to follow the project pattern and why it is good" and not just asking "just follow the project pattern".

I think an argument could also be made about not even needing the attribute since it can be read from the `territories.csv` file and with a bit of parsing we could accomplish the same result.

But, for now, I think we should follow the project pattern, for the sake of making it easier to find and some potential scripting, refactoring or adding of automatized pipeline elements.

processing/data_collection/gazette/spiders/rn_natal.py
giuliocc: I agree that it is not very complex, but `rrule` is a nicer interface regarding generating a date range.

If we could agree that following coding patterns in a project is good and we asked ourselves "what should be the coding pattern for date ranges in this project?", do you think that we should not use `rrule` and use `range`?

giuliocc:
No problems, if a comment is just providing noise it should not be made :) And I agree with you that a simpler spider is generally better than a complex one. It is good as it is, nicely done!

processing/data_collection/gazette/spiders/rn_natal.py
victor-torres: Moved to a constant.

processing/data_collection/gazette/spiders/rn_natal.py
victor-torres: Indeed, it's a nice interface.

But the question "should we make it a code pattern in this project?" should be addressed in another place as it's out of scope here. Perhaps we should raise an issue, tag it as a Question, and discuss the pros and cons there. For now, this project has 300+ spiders and it seems like only 3 of them are making use of `rrule` and only 2 of them are making use of the `AVAILABLE_FROM` constant. 

Questions that should be addressed before making it a pattern:

- how are we dealing with datetime imports?
  - `import datetime`
  - `import datetime as dt`
  - `from datetime import date, datetime`
- will `rrate` make it more difficult for first-time contributors or beginners?

giuliocc: Those are good points. Would you open the issue regarding this questions?

Just a correction: the project has 300+ spiders but most of Santa Catarina are based on the `FecamGazetteSpider`. We are talking about a 30-40 spiders project.

I am pointing this out because I think now is a good time to be raising this type of questions and defining some kind of guideline/style before it gets too hard to refactor.

I will consider this comment out of scope and approve this PR.

giuliocc:
@jvanz I approved the PR.

About the commits, I don't have write permissions so I leave it up to you to decide the best course of action :)

jvanz:
Thanks for your first contribution @victor-torres ! :)

processing/data_collection/gazette/spiders/rn_natal.py
victor-torres: https://github.com/okfn-brasil/diario-oficial/issues/193

PR #195, opened 2020-07-10
Add SIGPub (diariomunicipal.com.br) base spider and subspiders

giuliocc:
AMUPE (Associação de Municípios de Pernambuco) has some siblings :) The base is the same and it has coverage in almost every state (as you can see [here](http://www.diariomunicipal.com.br/)). Some states have most of their cities in it, some only have a bit more than a couple. But it is something already.

Right now, I implemented all states of the Northeast Region. I'll keep updating it in the next days. And 
document whatever we feel necessary at the end.

I reeeally didn't want to update CITIES.md to include these cities. Do you have a script for that since you added Fecam, @jvanz :smile:?

And `TERRITORY_ID` is kind of a problem. Because the Gazettes are for every city in the association. I thought about allowing a list of ids and adding multiple entries in the database for every id in the list. But I'm not sure about this. What do you think?

### DIFF
--- a/processing/data_collection/gazette/spiders/base.py
+++ b/processing/data_collection/gazette/spiders/base.py
@@ -1,7 +1,10 @@
+import json
 import re
-from datetime import datetime
+from datetime import date, datetime
 
 import dateparser
+from dateutil.rrule import DAILY, rrule
+from fake_useragent import UserAgent
 import scrapy
 from gazette.items import Gazette
 from scrapy.exceptions import NotConfigured
@@ -25,6 +28,87 @@ class BaseGazetteSpider(scrapy.Spider):
                 raise
         else:
             self.logger.info("Collecting all gazettes available")
+
+
+class SigpubGazetteSpider(BaseGazetteSpider):
+    """www.diariomunicipal.com.br (Sigpub) base spider
+
+    Documents obtained by this kind of spider are text-PDFs with many cities in it.
+    That's because the websites are usually made for associations of cities.
+    
+    TODO:
+        - All variations have a "possible" start date of 01/01/2009, but that may cause
+        many unnecessary requests to be made if they actually start making available
+        documents later. Some investigation for the start date of each website needs to
+        be made in this case.
+
+    Observations:
+        - These websites have an "Advanced Search", but they are protected by ReCaptcha.
+    """
+
+    custom_settings = {"USER_AGENT": UserAgent().random}
+    start_date = date(2009, 1, 1)
+
+    def start_requests(self):
+        """Requests start page where the calendar widget is available."""
+        yield scrapy.Request(self.CALENDAR_URL, callback=self.parse_calendar)
+
+    def parse_calendar(self, response):
+        """Makes requests for each date to see if a document is available."""
+        default_form_fields = {
+            "calendar[_token]": response.xpath(
+                "//input[@id='calendar__token']/@value"
+            ).get()
+        }
+        for date, date_form_fields in self.available_dates_form_fields():
+            formdata = {**default_form_fields, **date_form_fields}
+
+            yield scrapy.FormRequest(
+                url=response.urljoin("materia/calendario"),
+                formdata=formdata,
+                meta={"date": date, "edition_type": "regular"},
+                callback=self.parse_gazette_info,
+            )
+            yield scrapy.FormRequest(
+                url=response.urljoin("materia/calendario/extra"),
+                formdata=formdata,
+                meta={"date": date, "edition_type": "extra"},
+                callback=self.parse_gazette_info,
+            )
+
+    def parse_gazette_info(self, response):
+        """Parses document availability endpoint and gets document URL if available."""
+        body = json.loads(response.text)
+        meta = response.meta
+
+        if "error" in body:
+            self.logger.debug(
+                f"{meta['edition_type'].capitalize()} Gazette not available for {meta['date'].date()}"
+            )
+            return
+
+        for edition in body["edicao"]:
+            url = f"{body['url_arquivos']}{edition['link_diario']}.pdf"
+            yield Gazette(
+                date=meta["date"].date(),
+                file_urls=[url],
+                territory_id=self.TERRITORY_ID,
+                power="executive_legislative",
+                is_extra_edition=(meta["edition_type"] == "extra"),
+                scraped_at=datetime.utcnow(),
+                edition_number=edition.get("numero_edicao", ""),
+            )
+
+    def available_dates_form_fields(self):
+        """Generates dates and corresponding form fields for availability endpoint."""
+        available_dates = rrule(freq=DAILY, dtstart=self.start_date, until=date.today())
+        for query_date in available_dates:
+            form_fields = {
+                "calendar[day]": str(query_date.day),
+                "calendar[month]": str(query_date.month),
+                "calendar[year]": str(query_date.year),
+            }
+            yield query_date, form_fields
 
 
 class FecamGazetteSpider(BaseGazetteSpider):
--- a/processing/data_collection/gazette/spiders/ce_associacao_municipios.py
+++ b/processing/data_collection/gazette/spiders/ce_associacao_municipios.py
@@ -1,0 +1,7 @@
+from gazette.spiders.base import SigpubGazetteSpider
+
+
+class CeAssociacaoMunicipiosSpider(SigpubGazetteSpider):
+    name = "ce_associacao_municipios"
+    TERRITORY_ID = "2300000"
+    CALENDAR_URL = "http://www.diariomunicipal.com.br/aprece"
--- a/CITIES.md
+++ b/CITIES.md
@@ -414,7 +414,1865 @@ The municipality id (IBGE code) can be found on [Wikipedia](https://pt.wikipedia
 | 404 | Sarutaiá | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/232) |
 | 405 | Vera Cruz (RS) | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/232) |
 | 406 | Votorantim | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/232) |
-| 407 | Presidente Prudente | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/233) |
+| 407 | Abadia Dos Dourados - MG | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 408 | Abadia - GO | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 409 | Abadiania - GO | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 410 | Abaetetuba - PA | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 411 | Abaiara - CE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 412 | Abatiá - PR | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 413 | Abel Figueiredo - PA | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 414 | Abreu E Lima - PE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 415 | Acaiaca - MG | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 416 | Acari - RN | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 417 | Acopiara - CE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 418 | Acorizal - MT | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 419 | Acreúna - GO | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 420 | Afogados Da Ingazeira - PE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 421 | Afonso Bezerra - RN | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 422 | Afrânio - PE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 423 | Agrestina - PE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 424 | Água Azul Do Norte - PA | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 425 | Água Boa - MT | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 426 | Água Branca - PB | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 427 | Água Branca - PI | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 428 | Água Clara - MS | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 429 | Água Comprida - MG | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 430 | Água Fria - GO | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 431 | Água Nova - RN | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 432 | Água Preta - PE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 433 | Águas Belas - PE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 434 | Águas Vermelhas - MG | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 435 | Agudos Do Sul - PR | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 436 | Alagoa Nova - PB | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 437 | Alagoa - MG | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 438 | Alagoinha - PE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 439 | Alcântaras - CE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 440 | Alecrim - RS | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 441 | Alegrete - RS | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 442 | Alegria - RS | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 443 | Alexandria - RN | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 444 | Alexânia - GO | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 445 | Alhandra - PB | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 446 | Aliança - PE | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
+| 447 | Almadina - BA | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/195) |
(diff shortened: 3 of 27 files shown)


jvanz:
> AMUPE (Associação de Municípios de Pernambuco) has some siblings :) The base is the same and it has coverage in almost every state (as you can see [here](http://www.diariomunicipal.com.br/)). Some states have most of their cities in it, some only have a bit more than a couple. But it is something already.
> 
> Right now, I implemented all states of the Northeast Region. I'll keep updating it in the next days. And
> document whatever we feel necessary at the end.

Nice!

> I reeeally didn't want to update CITIES.md to include these cities. Do you have a script for that since you added Fecam, @jvanz smile?

I wrote a python script to generate the python files (jinja2) and use vim (with some macros) to update the CITIES.md file. You can do something similar using the data from the [territories.csv](https://github.com/okfn-brasil/diario-oficial/blob/master/data/territories.csv)

> And `TERRITORY_ID` is kind of a problem. Because the Gazettes are for every city in the association. I thought about allowing a list of ids and adding multiple entries in the database for every id in the list. But I'm not sure about this. What do you think?

I did not understand what you do mean. But why not do something similar to the FECAM spider? At the first glance, I think we can cover all the cities from all the states cover in that [website](http://www.diariomunicipal.com.br) with a single spider. We just need to has some control data in the subclasses of the same spider. 

Other question, who is responsible to maintain [this](http://www.diariomunicipal.com.br)?

giuliocc:
> I wrote a python script to generate the python files (jinja2) and use vim (with some macros) to update the CITIES.md file. You can do something similar using the data from the [territories.csv](https://github.com/okfn-brasil/diario-oficial/blob/master/data/territories.csv)

Nice. I'll make something then.

> I did not understand what you do mean. But why not do something similar to the FECAM spider? At the first glance, I think we can cover all the cities from all the states cover in that [website](http://www.diariomunicipal.com.br) with a single spider. We just need to has some control data in the subclasses of the same spider. 

Every scraped file is assigned to the generic `TERRITORY_ID` of the state (XX00000) right now because every gazette has sections for each city in the association. I would like to do something like Fecam, but that would result in duplicate files when downloaded from multiple spiders (in this case, one spider per city).

I was thinking of assigning the TERRITORY_ID of every city to the spider for the association (in a list), and when the file is being processed, instead of adding just one entry in the db with the generic TERRITORY_ID (as of now), we could add an entry for every TERRITORY_ID we assigned.

That way we could preserve the knowledge of what files correspond to what territories in the db and not make any duplicate requests. It would limit the number of spiders to a minimum which would be easier to maintain too.

> Other question, who is responsible to maintain [this](http://www.diariomunicipal.com.br)?
> 

[SIGPub is made by Vox Tecnologia](http://www.voxtecnologia.com.br/solucao-sigpub), which is a contractor for each association of cities.

jvanz:
> Every scraped file is assigned to the generic `TERRITORY_ID` of the state (XX00000) right now because every gazette has sections for each city in the association. I would like to do something like Fecam, but that would result in duplicate files when downloaded from multiple spiders (in this case, one spider per city).
> 
> I was thinking of assigning the TERRITORY_ID of every city to the spider of the association of cities (in a list), and when the file is being processed, instead of adding just one entry in the db with the generic TERRITORY_ID (as of now), we could add an entry for every TERRITORY_ID assigned to the spider.
> 
> That way we could preserve the knowledge of what files correspond to what territories in the db and not make any duplicate requests. It would limit the number of spiders to a minimum which would be easier to maintain too.

Hmm, one idea that just come to my mind. I did not validate it studying the web site... maybe I will try a POC.

Cannot you define which section should be processed in the subclass? In the base class, when you reach the point to decide which section should be processed, the base class call a method implemented by the subclasses. This method will return the section just for that specific city. Thus,  based on the return of that method, the base class can continue the normal process of find the files.

giuliocc:
> Cannot you define which section should be processed in the subclass? In the base class, when you reach the point to decide which section should be processed, the base class call a method implemented by the subclasses. This method will return the section just for that specific city. Thus, based on the return of that method, the base class can continue the normal process of find the files.

I'll work on something for it. A parser would be nice. I'll make updates on any advance.

jvanz:
But what do you think about this idea? Make sense to you?


‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

giuliocc:
It does :) You are talking about sections in the pdf, right? We could implement a generic parser that could divide the pdf into sections correctly and then each subclass could implement some configurations for the parser (like beginning and ending markers or regexes for the corresponding section). Is that similar to what you were thinking?

---

Juuust in case it really is:

I've looked into some stuff when talking to @vitorbaptista about implementing a parser for the gazettes.  When discussing it, we saw Google's Document AI and thought it was possible to do something like that here.

Although the logical approach to this would be a simpler approach first (like regex + `poppler-utils`)... it's just speculation, but I'm inclined to think this can get pretty complex pretty fast with the variety of document layouts we will be dealing with.

I'm currently looking into Document Layout Analysis to see what can be helpful. Stuff like [layout-parser](https://github.com/Layout-Parser/layout-parser) and [tesseract](https://github.com/madmaze/pytesseract) are already available. `poppler-utils`'s `pdftohtml` can be helpful too.

From my current understanding, a layout+text approach can be more generalizable and the more info we can get from the layout, the better. Like difference between fonts, bold text, location in page, separation of paragraphs, colored background, etc.

What do you think?

jvanz:
> It does :) You are talking about sections in the pdf, right? We could implement a generic parser that could divide the pdf into sections correctly and then each subclass could implement some configurations for the parser (like beginning and ending markers or regexes for the corresponding section). Is that similar to what you were thinking?

Actually, I'm talking about the spider to download the files only. I'm not referring to the pdf parser. 
> 
> Juuust in case it really is:
> 
> I've looked into some stuff when talking to @vitorbaptista about implementing a parser for the gazettes. When discussing it, we saw Google's Document AI and thought it was possible to do something like that here.
> 
> Although the logical approach to this would be a simpler approach first (like regex + `poppler-utils`)... it's just speculation, but I'm inclined to think this can get pretty complex pretty fast with the variety of document layouts we will be dealing with.
> 
> I'm currently looking into Document Layout Analysis to see what can be helpful. Stuff like [layout-parser](https://github.com/Layout-Parser/layout-parser) and [tesseract](https://github.com/madmaze/pytesseract) are already available. `poppler-utils`'s `pdftohtml` can be helpful too.
> 
> From my current understanding, a layout+text approach can be more generalizable and the more info we can get from the layout, the better. Like difference between fonts, bold text, location in page, separation of paragraphs, colored background, etc.
> 
> What do you think?

I'm open to any suggestion. But I think you should open an issue to discuss that. Let keep on this PR discussions about the spider implementation for now. 

I did not spend to much time researching about this yet. But it seems a cool idea and deserve a dedicated discussion thread for it. That's why I believe this is not the best place to talk about this and a issue only for that seems to be a better place.

jvanz:
Back to the spider implementation... I was walking thru the websites and noticed the advanced search page. AFAICS, you can update your spiders structure to use that page and filter gazettes by cities. At the first glance, it seems plausible.

giuliocc:
Sorry for the delay :/

> Back to the spider implementation... I was walking thru the websites and noticed the advanced search page. AFAICS, you can update your spiders structure to use that page and filter gazettes by cities. At the first glance, it seems plausible. 

I've tried doing that but the advanced searches I tested used recaptcha. I don't know if there are any plans to subsidy captcha breaking services in production but as there is a way to get the content without breaking recaptcha I didn't pay much attention to it.

> I'm open to any suggestion. But I think you should open an issue to discuss that. Let keep on this PR discussions about the spider implementation for now. 
> 
> I did not spend to much time researching about this yet. But it seems a cool idea and deserve a dedicated discussion thread for it. That's why I believe this is not the best place to talk about this and a issue only for that seems to be a better place. 

Nice. I am finishing the prototype of a parser to enable the text extraction of this spider. I think I will open the discussion when I finish the prototype and showcase on some spiders (maybe this one). Then we will have something more solid to discuss about. Is that ok?

gabriel19913:
Hi @giuliocc I'm following this discussion because other MG cities use a similar website. Just to make sure, for the cities in http://www.diariomunicipal.com.br/amm-mg/ I could implement a spider following the same logic and use the base class proposed at commit https://github.com/okfn-brasil/diario-oficial/pull/195/commits/2982cfb08b2850732270051ea94e53934743d2f9, right?
If I understand the discussion, those spiders will download the pdf's on the websites and the parser that you are doing will be responsible to get the information for each city from them, right?

giuliocc:
Hi @gabriel19913 !

> Just to make sure, for the cities in http://www.diariomunicipal.com.br/amm-mg/ I could implement a spider following the same logic and use the base class proposed at commit https://github.com/okfn-brasil/diario-oficial/pull/195/commits/2982cfb08b2850732270051ea94e53934743d2f9, right?

Yes, that's the way to do it right now. As of now I'm not adding any spiders for associations anymore since @jvanz suggested it would be better to break those up into spiders for each city and adding more of these wouldn't guarantee they would be used after all. But feel free to use my branch if you want to.

> If I understand the discussion, those spiders will download the pdf's on the websites and the parser that you are doing will be responsible to get the information for each city from them, right?

Yep. Althought that's just my current attempt into solving it. It's just an idea and I don't know if it will be accepted :) As suggested, a parser should be discussed in detail.

Thanks for joining. Hope we can work on it together :)

gabriel19913:
> Yes, that's the way to do it right now. As of now I'm not adding any spiders for associations anymore since @jvanz suggested it would be better to break those up into spiders for each city and adding more of these wouldn't guarantee they would be used after all. But feel free to use my branch if you want to.

Nice, I got it! I was thinking to make a spider for MG associations, but as you said it would be better to make for each city. However I don't know how this would be possible in this case.

> Yep. Althought that's just my current attempt into solving it. It's just an idea and I don't know if it will be accepted :) As suggested, a parser should be discussed in detail.

Nice! I think this will be a long discussion once working with pdf is not a simple task rsrs. But, if we could make something like that a lot of cities could be covered.

> Thanks for joining. Hope we can work on it together :)

Thank you. This is a really nice project and I hope I can contribute for it.

giuliocc:
Hey, do you think that it would be better to finish these spiders for the associations like it was being done at the beginning and then do a separate PR trying to break them up into spiders for each city? I think it should make this easier to review at least :/

Since we already have the `es_associacao_municipios` spider I think this wouldn't go against the projects patterns too much and we could unblock the data from a lot of cities (although not well structured).

processing/data_collection/gazette/spiders/base.py
```diff
@@ -18,6 +20,63 @@ def __init__(self, start_date=None, *args, **kwargs):
...
+                "regular": self.GAZETTE_INFO_URL,
+                "extra": self.EXTRA_GAZETTE_INFO_URL,
+            }
+            for edition_type, url in editions.items():
+                yield scrapy.FormRequest(
+                    url=url,
+                    formdata=formdata,
+                    meta={"date": query_date, "edition_type": edition_type},
+                    callback=self.parse_gazette_info,
+                )
```
jvanz, line 56:
```suggestion
        yield scrapy.FormRequest(
                    url= self.GAZETTE_INFO_URL,
                    formdata=formdata,
                    meta={"date": query_date, "edition_type": "regular"},
                    callback=self.parse_gazette_info,
        )
        yield scrapy.FormRequest(
                    url=self.EXTRA_GAZETTE_INFO_URL,
                    formdata=formdata,
                    meta={"date": query_date, "edition_type": "extra"},
                    callback=self.parse_gazette_info,
        )
```

Less indentations and loop control structures. ;)

jvanz (changes requested):
Please, can we split this PR in two? Keep only the spider on this one. And let's discuss the parser in another one. AFAiCS, this will not break anything. Right?

giuliocc:
I'll do it :) Thanks for the feedback

processing/data_collection/gazette/spiders/base.py
```diff
@@ -18,6 +20,63 @@ def __init__(self, start_date=None, *args, **kwargs):
...
+        yield scrapy.Request(self.CALENDAR_URL, callback=self.parse_calendar)
+
+    def parse_calendar(self, response):
+        formdata = {
+            "calendar[_token]": response.xpath(
+                "//input[@id='calendar__token']/@value"
+            ).get()
+        }
+        available_dates = rrule(
+            freq=DAILY, dtstart=self.FIRST_AVAILABLE_DATE, until=date.today()
```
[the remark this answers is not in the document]
giuliocc: Done

processing/data_collection/gazette/spiders/base.py
```diff
@@ -18,6 +20,63 @@ def __init__(self, start_date=None, *args, **kwargs):
...
+        )
+        for query_date in available_dates:
+            formdata.update(
+                {
+                    "calendar[day]": str(query_date.day),
+                    "calendar[month]": str(query_date.month),
+                    "calendar[year]": str(query_date.year),
+                }
+            )
```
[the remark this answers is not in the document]
giuliocc: Done

processing/data_collection/gazette/spiders/base.py
giuliocc: Done :)

giuliocc:
Everything all right. Removed the parser. Added the remaining spiders the way the first ones were made. Added some docstrings. Aaand updated `CITIES.md` (not fun, not fun).

Sorry for keeping you waiting.

alexandrevicenzi:
I want to raise one concern here as it was pointed out in previous comments but I don't think it was given proper attention.

The current approach used will download the same PDF for every city in a given association. I was looking to implement for Roraraima cities and I found that many other cities are using [/www.diariomunicipal.com.br/][http://www.diariomunicipal.com.br/].

While digging here and there I found that the PDF being download is one that contains multiple cities, so, to work properly the PDF should be split for each city, but this code just downloads the same PDF AFAIK.

The advanced search mentioned by @jvanz would be the proper way to download, this one returns a PDF per city, not association, in this case, we don't need to parse the PDF and extract anything because there's only once city, but the problem is Google Recaptcha.

Google Recaptcha pops randomly, I'm not sure if the current approach would work every time or if it would fail if running from a single IP.

The ideal way would be to break Recaptcha and parse for each city, otherwise, we will download the same PDF for many cities in some states and this does not really reflect the desired approach I think, but someone could confirm that for me.

I also noticed that some cities even registered in SIGPub do not seem to publish any Diarios.

My final questions are:

- Do we want to download the same exact PDF for all cities in the association multiple times?
- Do we want to parse data for one city and get data from another in it?

giuliocc:
Hey @alexandrevicenzi thanks for the comment! :)

> Google Recaptcha pops randomly, I'm not sure if the current approach would work every time or if it would fail if running from a single IP.

Although for the browser user ReCaptcha only pops up sometimes, in a crawler, the token has to be sent (or "solved") every time. Based on 2captcha's pricing ($3 for 1000 requests), a rough estimate for the entire database would be: `(3 usd * 12 years * 365 days * 1860 cities) / 1000 requests`.

To be more precise in this calculations we could account for days that don't have gazettes, splitting that number by a factor of, I don't know, 3 or 4? Also taking into account that the "Advanced Search" splits the gazettes into many files according to the "Órgão" field, the number should increase.

Based on this, I would expect this amount to be something around 10000 USD.

Also, using "Advanced Search" doesn't come without development costs because the files are split by "Órgão". We would have to accomodate that too. Is it worth it?

> Do we want to download the same exact PDF for all cities in the association multiple times?

That's a serious concern and I think that downloading only once is best to not put too much burden in the source website and in our storage. Right now with only a single spider for each association (or website).

> Do we want to parse data for one city and get data from another in it?

I'm not sure if I understood your question. But the idea is that later we should be able to split the content inside the PDFs by city. That way we wouldn't have content leaking between cities.

alexandrevicenzi:
> Based on this, I would expect this amount to be something around 10000 USD.

I don't expect to break Recaptcha, I just think that we need to document properly the behavior and limitations of such implementation.

> I'm not sure if I understood your question.

The current code does not seem to split the data correctly AFAIK, but it also does not parse for each city, so not a big deal.

The problem would be if each city has it's own parser and they download the same file.

processing/data_collection/gazette/spiders/base.py
```diff
@@ -18,6 +20,88 @@ def __init__(self, start_date=None, *args, **kwargs):
                 self.start_date = parsed_data.date()
 
 
+class DiarioMunicipalSigpubGazetteSpider(BaseGazetteSpider):
```
alexandrevicenzi, line 23: Perhaps just `SigPubGazetteSpider` as the class name would be shorter and still meaningful.

processing/data_collection/gazette/spiders/base.py
```diff
@@ -18,6 +20,88 @@ def __init__(self, start_date=None, *args, **kwargs):
...
+    Observations:
+        - These websites have an "Advanced Search", but they are protected by ReCaptcha.
+    """
+
+    custom_settings = {"USER_AGENT": UserAgent().random}
+    FIRST_AVAILABLE_DATE = date(2009, 1, 1)
+
+    def start_requests(self):
+        """Requests start page where the calendar widget is available."""
+        yield scrapy.Request(self.CALENDAR_URL, callback=self.parse_calendar)
```
alexandrevicenzi, line 44: The URLs are all the same, the only change in it is the association ID, it would look cleaner in the spiders that are using this base if they just pass association ID.

Example `url = f"{self.BASE_URL}/{self.ASSOCIATION_ID}/materia/calendario"` instead of `CALENDAR_URL`

And set in base class:

`BASE_URL = "http://www.diariomunicipal.com.br"`

And for each spider set:

`ASSOCIATION_ID = "somename"`

processing/data_collection/gazette/spiders/ce_associacao_municipios.py
```diff
@@ -0,0 +1,14 @@
+from datetime import date
+
+from gazette.spiders.base import DiarioMunicipalSigpubGazetteSpider
+
+
+class CeAssociacaoMunicipiosSpider(DiarioMunicipalSigpubGazetteSpider):
+    name = "ce_associacao_municipios"
+    TERRITORY_ID = "2300000"
+
+    CALENDAR_URL = "http://www.diariomunicipal.com.br/aprece"
```
alexandrevicenzi, line 10: As mentioned before, we can cleanup this as it's all the same. The only required bit is `aprece`.

giuliocc:
> I don't expect to break Recaptcha, I just think that we need to document properly the behavior and limitations of such implementation.

What would you suggest?

> The current code does not seem to split the data correctly AFAIK, but it also does not parse for each city, so not a big deal.
>
> The problem would be if each city has it's own parser and they download the same file.

Agreed. Based on #223 , data processing is being moved to another repo. So, we will be able to do this after downloading all necessary documents.

processing/data_collection/gazette/spiders/base.py
giuliocc: I wrote it that way because the name Sigpub is kind of hidden, so DiarioMunicipal could carry more meaning as it is the name of the website. But the name of the website is kind of generic then concatenating both seemed a good approach. But it is kind of big, yeah :/

I'll change it.

processing/data_collection/gazette/spiders/base.py
giuliocc: Just some context:

When I was doing this I didn't know how every website would behave since I did them one at a time. So separating every endpoint to it's own variable seemed safe. At the end, the pattern holds for every one of them.

I agree we could repurpose some parts of the URL. But I think that at least `CALENDAR_URL` shouldn't be split for documentation purposes. It points directly to the spider website and having the entire URL is nice.

I'll repurpose the other ones. Thanks!

alexandrevicenzi: Yeah, it does make sense, but the one from SC follows the approach I mentioned, still, looks cleaner than before.

processing/data_collection/gazette/spiders/base.py
alexandrevicenzi: The company behind it is Vox, this could also be in the name. DiarioMunicipal can be used by many states or even a city, too generic I guess.

giuliocc: Feels a bit weird putting the company name, doesn't it? hehehhee

alexandrevicenzi:
> What would you suggest?

Create a Wiki or documentation that address limitations, behaviors, why decisions were made, and everything else, so once someone joins the project they would have a place to understand why the project is the way it is and not file issues of weird things because they do not understand.

I proposed a change in #262, I would like to document a bit more in there what is the state of crawlers. I hope to get some feedback on how to do it better if not enough already.

giuliocc:
> Create a Wiki or documentation that address limitations, behaviors, why decisions were made, and everything else, so once someone joins the project they would have a place to understand why the project is the way it is and not file issues of weird things because they do not understand.

I like this and think it is worthy of debate. Maybe you could open up an issue on this topic. For this spider, I mentioned the ReCaptcha in the spider's docstring but I was brief to not make the docstring huge. Static documentation would help elaborating more on these topics.

processing/data_collection/gazette/spiders/base.py
```diff
@@ -24,6 +27,88 @@ def __init__(self, start_date=None, *args, **kwargs):
...
+    def parse_gazette_info(self, response):
+        """Parses document availability endpoint and gets document URL if available."""
+        body = json.loads(response.text)
+        meta = response.meta
+
+        if "error" in body:
+            self.logger.warning(
+                f"{meta['edition_type'].capitalize()} Gazette not available for {meta['date'].date()}"
+            )
+            return
```
jvanz, line 85: I do not think this worth a warning message. To me, it is more like a debug message. Just because a type of edition is not present does not mean there is something wrong or might be something wrong. Right? What do you think?

giuliocc: You're right. Changed it.

processing/data_collection/gazette/spiders/base.py
```diff
@@ -27,6 +30,88 @@ def __init__(self, start_date=None, *args, **kwargs):
...
+        many unnecessary requests to be made if they actually start making available
+        documents later. Some investigation for the start date of each website needs to
+        be made in this case.
+
+    Observations:
+        - These websites have an "Advanced Search", but they are protected by ReCaptcha.
+    """
+
+    custom_settings = {"USER_AGENT": UserAgent().random}
+    FIRST_AVAILABLE_DATE = date(2009, 1, 1)
```
jvanz, line 50: Can we use the `start_date` from the base class?

jvanz (changes requested):
Almost there! I promise! ;)

processing/data_collection/gazette/spiders/base.py
giuliocc: Nice catch! Done!

giuliocc:
No worries. Added `edition_number` capture as well :)

jvanz (approved):

giuliocc:
Just formatted with black to comply with Travis and fixed my pre-commit to not have this problem anymore :)

jvanz:
Thanks for your patience!

external-content duckduckgo com

PR #196, opened 2020-07-10
Bauru Gazette

Laerte:
I know that is already exist a PR for this, but because of the PR is 2 years old and is not merged, i resolve to implement the spider myself.


Laerte:
> Can you double check if the spider is working in your environment?
> 
> I ran it in my machine and I'm getting some errors:
> 
> ```
> 2020-07-14 02:02:28 [scrapy.core.scraper] ERROR: Error processing {'date': datetime.date(2009, 9, 29),                                                                                                             
>  'file_urls': ['http://www.bauru.sp.gov.br/arquivos/sist_diariooficial/2009/09/do_20090901_1710.pdf'],                                                                                                             
>  'files': [],                                                                                                                                                                                                      
>  'is_extra_edition': False,
>  'power': 'executive',
>  'scraped_at': datetime.datetime(2020, 7, 14, 2, 2, 19, 427232),
>  'territory_id': '3506003'}
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
>     current.result = callback(current.result, *args, **kw)
>   File "/mnt/code/data_collection/gazette/pipelines.py", line 60, in process_item
>     if self.is_doc(item["files"][0]["path"]):
> IndexError: list index out of range
> 2020-07-14 02:02:28 [scrapy.core.scraper] ERROR: Error processing {'date': datetime.date(2009, 9, 29),
>  'file_urls': ['http://www.bauru.sp.gov.br/arquivos/sist_diariooficial/2009/09/do_20090903_1711.pdf'],
>  'files': [],
>  'is_extra_edition': False,
>  'power': 'executive',
>  'scraped_at': datetime.datetime(2020, 7, 14, 2, 2, 19, 426462),
>  'territory_id': '3506003'}
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
>     current.result = callback(current.result, *args, **kw)
>   File "/mnt/code/data_collection/gazette/pipelines.py", line 60, in process_item
>     if self.is_doc(item["files"][0]["path"]):
> IndexError: list index out of range
> ```
> 
> What's strange to me is that if I open the URL from `file_urls` field, I can see the gazette. Can you just rerun and check if everything is fine? I used the following command to run your spider:
> 
> `SPIDER=sp_bauru make run_spider`

Hi! I facing this problem too and its in **ExtractTextPipeline**:

debug

jvanz:
> For some reason `files` not exists... when you run anothers spiders you don't get any errors?

No, I don't. I just run another spider (`sc_gaspar`) from your branch and everything worked as expected. My bet is that redirecting (302) is causing issues. But I could not spend some time to investigate it. Maybe there is some kind of settings to enable redirecting.

Laerte:
> > For some reason `files` not exists... when you run anothers spiders you don't get any errors?
> 
> No, I don't. I just run another spider (`sc_gaspar`) from your branch and everything worked as expected. My bet is that redirecting (302) is causing issues. But I could not spend some time to investigate it. Maybe there is some kind of settings to enable redirecting.

Yes! That's right. I enable redirect for media pipelines, by [default](https://doc.scrapy.org/en/latest/topics/media-pipeline.html#allowing-redirections) is disabled, do you mind test again?

processing/data_collection/gazette/settings.py
```diff
@@ -13,3 +13,4 @@
 # skip already crawled item
 DELTAFETCH_ENABLED = True
 DELTAFETCH_DIR = "/mnt/data/deltafetch"
+MEDIA_ALLOW_REDIRECTS = True
```
jvanz, line 16: Thanks for your research! But I would like to ask you one last thing. Move this configuration for the spider. You can see how to do it [here](https://docs.scrapy.org/en/latest/topics/settings.html#settings-per-spider). As most of the spider work without this, I would like to keep this confirm just for this spider. Later, when we face more spider with this issue, we can move the setting to all spider.

Laerte: No problem, done.

jvanz:
As @rodolfolottin mention in the #197 PR we do not use semantic commit message. But to avoid this ping-pong of rebase commits, I'll squash and merge your changes. Thus we can keep the history clean and concise. 

Once again. thank you very much for your contribution. You're helping make a better World! Hehe, quite modest statement, isn't it? :joy:

Issue #205, opened 2020-08-15
Add command to list all cities


fbidu:
As the project grows, it may be useful to have a command that lists all the available cities

```bash
python gazette list  # Lists all
python gazette list  --estado=sp # Lists all from SP
python gazette list  --estado=sp,rj,sc # Lists all from SP, RJ e SC
```

What do you think?

PR #206, opened 2020-08-15
Add List command

fbidu:
This is related to #205 and it adds a CLI to Gazette that lists all the available spiders. It accepts an optional list of states' initials and filters by them if they are provided. The output is grouped by the state's initials, in alphabetical order and the cities inside the group are also sorted.

I'm not very proficient with Python's `argparse` so I might have not done things in an optimal way, if anyone can take a closer look at argument handling, please do!

The expected behavior is:
Peek 15-08-2020 17-49

Please notice that I'm inside the `processing/data_collection` folder. Maybe it would be nice to have a way to externalize this command to the project's root

### DIFF
--- a/processing/data_collection/gazette/__main__.py
+++ b/processing/data_collection/gazette/__main__.py
@@ -1,0 +1,26 @@
+from collections import defaultdict
+import re
+from pathlib import Path
+
+
+def main():
+    root = Path(__file__).parent
+    spiders = root / "spiders"
+    pattern = r"([a-z]{2})_([a-z_]+)\.py"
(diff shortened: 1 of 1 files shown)


anapaulagomes (changes requested):
Nice one, @fbidu!

Could you please move the script to our `scripts` folder (https://github.com/okfn-brasil/querido-diario/tree/main/scripts) and update the README with info about this command? I'm gonna take a look as soon as you push it. :)

rennerocha:
Closing stale PR.

PR #208, opened 2020-08-15
Duplicate key when running make setup

jvanz:
One of the step ran by the "make setup" is to populate the territories
table with the content of the "territories.cvs" files. However, if the
database is already populates by some previous failed "make setup" run,
The command fails. To fix that, before start the process of build
container images and start the process, a docker-compose call is done to
remove all the container images and volume used by the services. Thus,
the "make setup" will always rebuild everything fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-setup:
+setup: destroy
 	cp .env.example .env
 	docker-compose pull
 	docker-compose build
@@ -9,6 +9,10 @@ setup:
 		pip install pre-commit; \
 	fi
 	pre-commit install
+
+destroy:
+	echo "Removing all containers...."
+	docker-compose down --rmi all --volumes --remove-orphans
 
 seed:
 	docker-compose up -d postgres


giuliocc:
@jvanz sorry I can't test it right now I will try it later. But from what I understood from the docs, `make destroy` would erase the images, containers and networks too, right? Do you think that in this case, when running `make setup` it should just clean the existing volumes instead? To avoid having to rebuild everything?

jvanz:
> @jvanz sorry I can't test it right now I will try it later. But from what I understood from the docs, `make destroy` would erase the images, containers and networks too, right? Do you think that in this case, when running `make setup` it should clean everything but leave the images instead? To avoid having to rebuild everything?

Hum, the goal of the `make setup` is to build the images. But I understand your point. I`ll add a config variable with the default value to not delete the images.

giuliocc:
Tested it. Worked well. Kudos! (it even doesn't delete the image if other containers use it)

PR #209, opened 2020-08-15
Change pre-commit hook

BrunaNayara:
The `make setup` installed `pre-commit` to create and maintain git hooks. But running it resulted in this error 
```pre-commit install
make: pre-commit: Command not found
Makefile:2: recipe for target 'setup' failed
make: *** [setup] Error 127
```
I created a small pre-commit hook, and added black to requirements, because the only hook was black*.

\* Refer to the [CONTRIBUTING](https://github.com/okfn-brasil/diario-oficial/blob/main/CONTRIBUTING.md#automated-code-formatting) and [pre commit config](https://github.com/okfn-brasil/diario-oficial/blob/main/.pre-commit-config.yaml) files


requirements.txt
```diff
@@ -0,0 +1 @@
+black==19.10b0
```
sergiomario, line 1: requirements.txt exists here: https://github.com/okfn-brasil/diario-oficial/blob/main/processing/requirements.txt

BrunaNayara: Oh, nice. I'll point to that one :)

BrunaNayara: Actually, this requirements is for the docker container and they are likely to conflict if installing directly out of docker.
So I don't think it's good installing them in the setup. They should only be installed in the docker steps. 
@sergiomario what do you think?

sergiomario: Right. I understood the conflict, I'll do the merge like this.

jvanz:
To keep the tradition of cool gif when a new contributor send the first PR:
image

jvanz:
giphy

PR #210, opened 2020-08-15
Add Spider for Valinhos/SP

bbertolani:
Add spider for Valinhos

Thanks to Vans, Giulio and all the other hosts of 1 Sprint Querido Diario


giuliocc:
Hey @bbertolani ! That's great! Thanks for the PR :) I will look better into it later to help the review process.

processing/data_collection/gazette/spiders/sp_valinhos.py
```diff
@@ -0,0 +1,45 @@
...
+        items = []
+        blocks = response.xpath('//*[@id="block-system-main"]/div/div[2]/div/ul/li')
+        for edition in blocks:
+            span = edition.xpath("//div/span/div/div[2]/a")
+            url = span.css("::attr(href)").extract_first().replace("../", "")
+
+            spanDate = edition.xpath("//div/span/div/div[2]/span")
+            date = spanDate.css("::attr(content)").extract_first()
+
+            items.append(
```
rennerocha, line 37: You don't need to append all items and return them all together. This is a generator, so you could simple yield it inside this loop:

```
                yields Gazette(
                    date=date,
                    file_urls=[url],
                    territory_id=self.TERRITORY_ID,
                    scraped_at=datetime.datetime.utcnow(),
                )
```

processing/data_collection/gazette/spiders/sp_valinhos.py
```diff
@@ -0,0 +1,45 @@
...
+
+                url = self.selector_url.format(month, year)
+                yield scrapy.Request(url, self.parse_month_page)
+
+    def parse_month_page(self, response):
+        items = []
+        blocks = response.xpath('//*[@id="block-system-main"]/div/div[2]/div/ul/li')
+        for edition in blocks:
+            span = edition.xpath("//div/span/div/div[2]/a")
+            url = span.css("::attr(href)").extract_first().replace("../", "")
```
rennerocha, line 32: This is not critical, but Scrapy is favoring the use of `get` and `getall`, instead of `extract` and `extract_first` (https://docs.scrapy.org/en/latest/topics/selectors.html?highlight=extract_first#extract-and-extract-first).

processing/data_collection/gazette/spiders/sp_valinhos.py
```diff
@@ -0,0 +1,45 @@
...
+            for month in range(1, 13):
+                if year == today.year and month > today.month:
+                    return
+
+                url = self.selector_url.format(month, year)
+                yield scrapy.Request(url, self.parse_month_page)
+
+    def parse_month_page(self, response):
+        items = []
+        blocks = response.xpath('//*[@id="block-system-main"]/div/div[2]/div/ul/li')
```
rennerocha, line 29: As rule of thumb, it is better to have your selectors as simple as possible. This will make maintenance much easier.
If for some reason, a new `<div>` is added, it will will break your selector and you will stop to get the data, even if the page didn't change much.

If you look better at the source code of the page, you will see that what you really want are the `<li>` tags that are inside a `<div>` with class `item-list`.

A much simpler selector (you can use CSS selectors!) would be:
`response.css("div.item-list li")`

If you look better at the content and the classes of the page, you can do even more specific, getting only the `li` tags that contains `views-row` class:

`response.css("li.views-row")` or `response.css(".views-row")`

processing/data_collection/gazette/spiders/sp_valinhos.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
```
rennerocha, line 1: This is not necessary with Python 3.7. It could be removed.

processing/data_collection/gazette/spiders/sp_valinhos.py
rennerocha: Each edition block has only one `<a>`, so I didn't understood why you need to assign a `span` variable only to get the URL. The following line will do the same as these two lines and are much more simple to understand:

```
url edition.css("a::attr(href)").get()
```

processing/data_collection/gazette/spiders/sp_valinhos.py
```diff
@@ -0,0 +1,45 @@
...
+                yield scrapy.Request(url, self.parse_month_page)
+
+    def parse_month_page(self, response):
+        items = []
+        blocks = response.xpath('//*[@id="block-system-main"]/div/div[2]/div/ul/li')
+        for edition in blocks:
+            span = edition.xpath("//div/span/div/div[2]/a")
+            url = span.css("::attr(href)").extract_first().replace("../", "")
+
+            spanDate = edition.xpath("//div/span/div/div[2]/span")
```
rennerocha, line 34: When using XPath, you need to be careful. When your selector starts with `/`, it will search the entire document, not only the `edition` selector. This certainly will return the same `span` tag for every iteration of the loop.

The correct would be `".//div/span/div/div[2]/span"`, that returns the tags relative to `edition`.

However, it is not necessary this complicated selector to obtain the date. See my next comment.

processing/data_collection/gazette/spiders/sp_valinhos.py
```diff
@@ -0,0 +1,45 @@
...
+            date = spanDate.css("::attr(content)").extract_first()
```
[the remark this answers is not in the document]
bbertolani: Tks for all tips. I will work on it.

jaswdr:
@bbertolani any plans to continue this PR?

bbertolani:
oh sure, I will find time this week to work on it.

anapaulagomes:
@bbertolani let us know if you need any help!

rennerocha:
Completed in https://github.com/okfn-brasil/querido-diario/pull/598

Issue #182, opened 2020-06-20
Belém spider


jvanz:
Add an spider for Belém, capital of Pará

rodolfolottin:
I think that we can close this issue now 😄

PR #212, opened 2020-08-17
Add spider for Belem - Pará

rodolfolottin:
### Description

Add spider for Belem - Pará. Issue #182 .

I've tried to run the spider built on [PR 87](https://github.com/okfn-brasil/diario-oficial/pull/87) but it didn't work and based on the spider code I assumed that the website changed since it was built (2 years ago)

Despite that, this PR is still a WIP because I couldn't find a way to download the PDF files.

#### Problem description

On this spider [I am using a different URL](https://sistemas.belem.pa.gov.br/diario-consulta-api/diarios) (the API one) but you can [check in here](http://sistemas.belem.pa.gov.br/diario/painel) that we can't get any pdf `href` based on the website DOM. So, as we can download the file just by clicking on the button I guess that the only possible solution that we have is to install selenium, right? Do you see any other possible solution?

As I couldn't resolve it by scraping the DOM - because of Javascript - and I wanted to check if there is any other option than downloading selenium I've decided to try to use the API url and see if I can hack a way out to download the gazette pdfs, but unfortunately I couldn't.

If I try to download the file specifying the Accept headers key, like here: `curl -X GET https://sistemas.belem.pa.gov.br/diario-consulta-api/diarios/13959 -H 'Accept: application/octet-stream'` the file is downloaded, but if I don't specify it - that's what happens when we [`yield` the `Gazette`](https://github.com/okfn-brasil/diario-oficial/compare/main...rodolfolottin:pa_belem_spider?expand=1#diff-9682884629ccc69202e9125a3119036cR49-R56) the downloaded content is just a pure JSON.

What do you think? Do you see any other solution than installing selenium?

### DIFF
--- a/CITIES.md
+++ b/CITIES.md
@@ -19,7 +19,7 @@ The municipality id (IBGE code) can be found on [Wikipedia](https://pt.wikipedia
 | 9 | Recife | :white_check_mark: | [issue](https://github.com/okfn-brasil/diario-oficial/issues/187) | [PR](https://github.com/okfn-brasil/diario-oficial/pull/170) |
 | 10 | Porto Alegre | :white_check_mark: | | |
 | 11 | Goiânia | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/6) |
-| 12 | Belém | | | |
+| 12 | Belém | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/212) |
 | 13 | Guarulhos | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/4) |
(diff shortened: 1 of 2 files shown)


giuliocc:
Hey! Great spider and analysis of the problem.

Thankfully Scrapy already has a solution for this :) If you change the setting `DEFAULT_REQUEST_HEADERS` you can specify the 'Accept' header to include `application/octet-stream`.

I tested it adding this class attribute to the spider and the pdfs were correctly downloaded:
```python
    custom_settings = {
        'DEFAULT_REQUEST_HEADERS': {
            'Accept': 'text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8,application/octet-stream'
        }
    }
```

Hope this helps!

rodolfolottin:
You rock @giuliocc ! I didn't know about that

I think that now the PR is ready 😄

PR #224, opened 2020-08-23
Add spider for Federação dos Municípios do Rio Grande do Norte

andradejunior:
This PR adds a spider for FEMURN (Federação dos Municípios do Rio Grande do Norte) that brings data from a variety of municipalities in Rio Grande do Norte in only one gazette.

### DIFF
--- a/processing/data_collection/gazette/spiders/rn_federacao_municipios.py
+++ b/processing/data_collection/gazette/spiders/rn_federacao_municipios.py
@@ -1,0 +1,56 @@
+import json
+from dateparser import parse
+import datetime as dt
+
+import scrapy
+
+from gazette.items import Gazette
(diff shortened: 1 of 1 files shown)


giuliocc:
Hey, @andradejunior ! I realized this spider is already covered in #195 (I've changed the PR name now to make it easier to notice that). Maybe you could work with us there? Currently we are trying to solve the issue of being able to split the content by city. Since you already have some experience with the spider's system, your help would be immense!

vitorbaptista:
Assuming that #195 supersedes this PR, could we close it and move the discussion to #195?

jvanz:
Yes, I think we can close this one. The PR #195 cover this and more cases.

PR #229, opened 2020-09-01
Adding spider for Itauna,MG

pgarcias01:
This PR adds a spider for Itaúna/MG.

For being my first contribution I apologize for any mistakes.

### DIFF
--- a/processing/data_collection/gazette/spiders/mg_itauna.py
+++ b/processing/data_collection/gazette/spiders/mg_itauna.py
@@ -1,0 +1,36 @@
+import datetime as dt
+
+from dateparser import parse
+from scrapy import Request
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class MgItaunaSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3133808"
+    name = "mg_itauna"
+    allowed_domains = ["itauna.mg.gov.br"]
+    start_urls = ["https://www.itauna.mg.gov.br/portal/diario-oficial/"]
+
+    def parse(self, response):
+        diarios = response.xpath("//div[@class='d_e_modelo_diario']")
+        for diario in diarios:
+            url = (
+                "https://www.itauna.mg.gov.br"
+                + diario.css("a")[1].extract().split('"')[1]
+            )
+            date = diario.xpath("div/span/text()").extract()[0].split()[0]
+            is_extra_edition = True if len(diario.css("i")) == 4 else False
+            yield Gazette(
+                date=parse(date, languages=["pt"]).date(),
+                file_urls=[url],
+                is_extra_edition=is_extra_edition,
+                territory_id=self.TERRITORY_ID,
+                power="executive_legislature",
+                scraped_at=dt.datetime.utcnow(),
+            )
+        url = "https://www.itauna.mg.gov.br/portal/diario-oficial/{i}/0/0/0/0/"
+        last_page = int(response.xpath("//select/option/text()")[-1].extract())
+        for i in range(1, 1 + last_page):
+            yield Request(url.format(i=i), callback=self.parse)
--- a/processing/data_collection/gazette/spiders/mg_itauna.py
+++ b/processing/data_collection/gazette/spiders/mg_itauna.py
@@ -13,13 +13,15 @@ class MgItaunaSpider(BaseGazetteSpider):
     start_urls = ["https://www.itauna.mg.gov.br/portal/diario-oficial/"]
 
     def parse(self, response):
-        url = self.start_urls[0] + "{i}/0/0/0/0/"
-        last_page = int(response.xpath("//select/option/text()")[-1].extract())
-        for i in range(1, 1 + last_page):
-            yield Request(url.format(i=i), callback=self.parse_editions_page)
+        page_url = response.urljoin("{page}/0/0/0/0/")
+        last_page = int(
+            response.xpath("//select[@id='select']/option[last()]/text()").get()
+        )
+        for page in range(1, 1 + last_page):
+            yield Request(url.format(page=page), callback=self.parse_editions_page)
 
     def parse_editions_page(self, response):
-        diarios = response.xpath("//div[@class='d_e_modelo_diario']")
+        diarios = response.css(".d_e_modelo_diario")
         for diario in diarios:
             href = diario.xpath('.//a[contains(@href, "downloads")]/@href').get()
             date = diario.xpath("div/span/text()").re_first("\d{2}/\d{2}/\d{4}")
--- a/processing/data_collection/gazette/spiders/mg_itauna.py
+++ b/processing/data_collection/gazette/spiders/mg_itauna.py
@@ -18,7 +18,7 @@ class MgItaunaSpider(BaseGazetteSpider):
             response.xpath("//select[@id='select']/option[last()]/text()").get()
         )
         for page in range(1, 1 + last_page):
-            yield Request(url.format(page=page), callback=self.parse_editions_page)
+            yield Request(page_url.format(page=page), callback=self.parse_editions_page)
 
     def parse_editions_page(self, response):
         diarios = response.css(".d_e_modelo_diario")
--- a/processing/data_collection/gazette/spiders/mg_itauna.py
+++ b/processing/data_collection/gazette/spiders/mg_itauna.py
@@ -1,5 +1,4 @@
 import datetime as dt
-
 from dateparser import parse
 from scrapy import Request
 
@@ -14,23 +13,22 @@ class MgItaunaSpider(BaseGazetteSpider):
     start_urls = ["https://www.itauna.mg.gov.br/portal/diario-oficial/"]
 
     def parse(self, response):
+        url = self.start_urls[0] + "{i}/0/0/0/0/"
+        last_page = int(response.xpath("//select/option/text()")[-1].extract())
+        for i in range(1, 1 + last_page):
+            yield Request(url.format(i=i), callback=self.parse_editions_page)
+
+    def parse_editions_page(self, response):
         diarios = response.xpath("//div[@class='d_e_modelo_diario']")
         for diario in diarios:
-            url = (
-                "https://www.itauna.mg.gov.br"
-                + diario.css("a")[1].extract().split('"')[1]
-            )
-            date = diario.xpath("div/span/text()").extract()[0].split()[0]
-            is_extra_edition = True if len(diario.css("i")) == 4 else False
+            href = diario.xpath('.//a[contains(@href, "downloads")]/@href').get()
+            date = diario.xpath("div/span/text()").re_first("\d{2}/\d{2}/\d{4}")
+            is_extra_edition = diario.xpath(".//span[contains(./text(), 'Extra')]").get() is not None
             yield Gazette(
                 date=parse(date, languages=["pt"]).date(),
-                file_urls=[url],
+                file_urls=[response.urljoin(href)],
                 is_extra_edition=is_extra_edition,
                 territory_id=self.TERRITORY_ID,
                 power="executive_legislature",
                 scraped_at=dt.datetime.utcnow(),
             )
-        url = "https://www.itauna.mg.gov.br/portal/diario-oficial/{i}/0/0/0/0/"
-        last_page = int(response.xpath("//select/option/text()")[-1].extract())
-        for i in range(1, 1 + last_page):
-            yield Request(url.format(i=i), callback=self.parse)
--- a/processing/data_collection/gazette/spiders/mg_itauna.py
+++ b/processing/data_collection/gazette/spiders/mg_itauna.py
@@ -17,7 +17,7 @@ class MgItaunaSpider(BaseGazetteSpider):
         last_page = int(
             response.xpath("//select[@id='select']/option[last()]/text()").get()
         )
-        for page in range(1, 1 + last_page):
+        for page in range(2, 1 + last_page):
             yield Request(page_url.format(page=page), callback=self.parse_editions_page)
 
     def parse_editions_page(self, response):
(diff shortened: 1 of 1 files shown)


processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -0,0 +1,36 @@
...
+    name = "mg_itauna"
+    allowed_domains = ["itauna.mg.gov.br"]
+    start_urls = ["https://www.itauna.mg.gov.br/portal/diario-oficial/"]
+
+    def parse(self, response):
+        diarios = response.xpath("//div[@class='d_e_modelo_diario']")
+        for diario in diarios:
+            url = (
+                "https://www.itauna.mg.gov.br"
+                + diario.css("a")[1].extract().split('"')[1]
```
giuliocc, line 21: As a tip, these kinds of selections can break easily if the website adds one element `a` before the one you are expecting or just inverting the order of the links.

One thing you could do in this case is look for a pattern that these links have and use it.

An example is this xpath `.//a[contains(@href, "downloads")]/@href` which only retrieves the `href` attribute of `a` elements which contains the string "downloads" in it's `href` attribute.

giuliocc: As another tip, you can use `::attr()` (in CSS selectors) to specify which attribute you want to extract. That way you wouldn't need to split by `"` and select the first attribute.

giuliocc: Maybe using `response.urljoin(href)` instead of `"https://www.itauna.mg.gov.br" + href` could be better since you don't need to hardcode the base url?

processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -0,0 +1,36 @@
...
+            )
+            date = diario.xpath("div/span/text()").extract()[0].split()[0]
```
giuliocc, line 23: Newer versions of Scrapy are favoring `.getall()` as a replacement for `.extract()`.

giuliocc: Also, `.get()` is a nice shorcut for `.getall()[0]` :)

giuliocc: Maybe using `.re_first("\d{2}/\d{2}/\d{4}")` could avoid potential problems in the future if they decide to change the order of the fields?

processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -0,0 +1,36 @@
...
+            is_extra_edition = True if len(diario.css("i")) == 4 else False
```
giuliocc, line 24: For the same reasons as given before I think that this check could render better results since you would be checking for the a button that explicitly contains the string "Extra" (maybe "Edição Extra" if you feel like it):

`is_extra_edition = diario.xpath(".//span[contains(./text(), 'Extra')]").get() is not None`

processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -0,0 +1,36 @@
...
+                file_urls=[url],
+                is_extra_edition=is_extra_edition,
+                territory_id=self.TERRITORY_ID,
+                power="executive_legislature",
+                scraped_at=dt.datetime.utcnow(),
+            )
+        url = "https://www.itauna.mg.gov.br/portal/diario-oficial/{i}/0/0/0/0/"
+        last_page = int(response.xpath("//select/option/text()")[-1].extract())
+        for i in range(1, 1 + last_page):
+            yield Request(url.format(i=i), callback=self.parse)
```
giuliocc, line 36: Maybe we could separate the pagination from the actual parsing? This could avoid confusion due to the "recursiveness".

Like:

```python
...
def parse(self, response):
	url = "https://www.itauna.mg.gov.br/portal/diario-oficial/{i}/0/0/0/0/"
	for i in range(1, last_page + 1):
        yield Request(url.format(i=i), callback=self.parse_editions_page)

def parse_editions_page(self, response):
	diarios = response.xpath("//div[@class='d_e_modelo_diario']")
    for diario in diarios:
	...
...
```

pgarcias01:
> Hey @pgarcias01 ! Great spider overall, thanks for contributing :) Just gave some tips that could improve it since it is already working as expected.
> 
> Just a note for the future: some pdfs are images (25/08/2020), and some gazettes are text but have some images inside (01/09/2020)

Hi @giuliocc I made the changes as suggested, can you take a look again?

processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class MgItaunaSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3133808"
+    name = "mg_itauna"
+    allowed_domains = ["itauna.mg.gov.br"]
+    start_urls = ["https://www.itauna.mg.gov.br/portal/diario-oficial/"]
+
+    def parse(self, response):
+        url = self.start_urls[0] + "{i}/0/0/0/0/"
```
rennerocha, line 16: **suggestion**:
It is much safer to use `response.urljoin` to format new URLs instead of concatenating strings manually.
`url = response.urljoin("{i}/0/0/0/0/")` will produce the same result.

**nitpick**: The use of one character variables names decrease the readability of the code, as we don't know exactly what that variable contains. For example, `i` doesn't provide me context. `url = response.urljoin("{page}/0/0/0/0/")` shows the intent of that variable, so we know that every time we format a string, that value will referencing the page number that we want to retrieve.

**nitpick**: This is much more personal, so fell free to ignore as well (as all nitpicks can be ignored). I would prefer to name this variable as `page_url`, to make it clear that this is the URL of a page.

processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -0,0 +1,36 @@
...
+        last_page = int(response.xpath("//select/option/text()")[-1].extract())
```
rennerocha, line 17: **suggestion**: You can use `last()` as parameter to get the last `option` in the `select`. Doing that you avoid to 
`last_page = int(response.xpath("//select[@id='select']/option[last()]/text()").get())`

**nitpick**: Scrapy suggest that you use `get` (instead of `extract_first`) and `getall` (instead of `extract`) now. The old works, but it would be good if we keep consistency in the project.

processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -0,0 +1,36 @@
...
+        for i in range(1, 1 + last_page):
```
rennerocha, line 18: **suggestion**: As I mentioned before, a clear variable name would make the code more readable.

Compare this:
```
for i in range(1, 1 + last_page):
    yield Request(url.format(i=i, callback=self.parse_editions_page)
```

With this:
```
for page in range(1, 1 + last_page):
    yield Request(url.format(page=page), callback=self.parse_editions_page)
```

In my opinion the second is much more readable.

processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -0,0 +1,36 @@
...
+            yield Request(url.format(i=i), callback=self.parse_editions_page)
+
+    def parse_editions_page(self, response):
+        diarios = response.xpath("//div[@class='d_e_modelo_diario']")
```
rennerocha, line 22: **issue**:
It is a bit dangerous to search for a class using XPath like that. This will try to find a `div` where the class is **exactly** equal to `d_e_modelo_diario`. 

`response.xpath("//div[@class='d_e_modelo_diario']")` will match:

`<div class="d_e_modelo_diario"><!-- My DIV content --></div>`

But it will not match (look the extra whitespace):
`<div class="d_e_modelo_diario "><!-- My DIV content --></div>`

It will not match this one either:
`<div class="d_e_modelo_diario another_class"><!-- My DIV content --></div>`

So a single whitespace may break everything. For class search, I prefer to use a CSS Selector, that avoid this problem:

`diarios =response.css("div.d_e_modelo_diario")` or `diarios =response.css(".d_e_modelo_diario")` if this is the only tag with this calss.

processing/data_collection/gazette/spiders/mg_itauna.py
giuliocc, line 17: Is there a class or id that can narrow your search space and make you find this last page in a similar way to the others xpaths you are using? For the same reasons I gave before, if you are not sure that if they can introduce a new unrelated select element and brake this, it might be better trying to think another way of finding this element.

This xpath is very simple and that is very good, but if wasn't looking for every select element in the page it could avoid some problems.

processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -0,0 +1,36 @@
...
+    allowed_domains = ["itauna.mg.gov.br"]
+    start_urls = ["https://www.itauna.mg.gov.br/portal/diario-oficial/"]
+
+    def parse(self, response):
+        url = self.start_urls[0] + "{i}/0/0/0/0/"
+        last_page = int(response.xpath("//select/option/text()")[-1].extract())
+        for i in range(1, 1 + last_page):
+            yield Request(url.format(i=i), callback=self.parse_editions_page)
+
+    def parse_editions_page(self, response):
```
rennerocha, line 21: **thought**: The URL you defined in `start_urls` is the equivalent of `https://www.itauna.mg.gov.br/portal/diario-oficial/1/0/0/0/0/` (the first page of the results). Because of that, all the logic that are inside this method could be placed inside `parse` method, and the pagination could start on `2` instead of `1`. The way you did now, add one extra request to the whole scraping process. This is not too bad, but it is better to send as few requests as possible always. It is not a problem to reuse the same callback for processing a page and handling its pagination.

processing/data_collection/gazette/spiders/mg_itauna.py
rennerocha: `select` with `id="select"` is the one that we need to get. This is enough to get only the one that we are interested.

processing/data_collection/gazette/spiders/mg_itauna.py
rennerocha: This doesn't introduce confusion. It is a very well known pattern in spiders. The solution you proposed will also introduce an extra request to page `1` that could be avoided if we do everything in `parse` method. As the spider is short, it won't be a problem.

giuliocc: Then maybe adding a `yield from self.parse_editions_page(response)` at the end of the `parse` method and change the range to start at 2?

This addresses the unnecessary request maintaining separations of concerns.

I said it could create some confusion because the loop is creating new requests to the same method that creates those same requests. If the dupefilter wasn't in place this would be an infinite loop. And it would be better to address this in an explicit way than a implicit one.

Should have explained it a bit better.

processing/data_collection/gazette/spiders/mg_itauna.py
giuliocc: True. That was my mistake.

processing/data_collection/gazette/spiders/mg_itauna.py
pgarcias01: @rennerocha and @giuliocc , thanks for the patience and the tips, my idea is to continue with other municipalities here in MG. @rennerocha  in the first version had done it with only one function, but I believe that segmented like that, it was easier to understand.

giuliocc: Great, @pgarcias01 . The more hands the better ;)
---
I put this response in another comment made by Renne. Since Github's notifications always confuses me I don't know if you saw it or not so I'll just paste it here :)

> Then maybe adding a `yield from self.parse_editions_page(response)` at the end of the `parse` method and change the range to start at 2?
> 
> This addresses the unnecessary request maintaining separations of concerns.
> 
> I said it could create some confusion because the loop is creating new requests to the same method that creates those same requests. If the dupefilter wasn't in place this would be an infinite loop. And it would be better to address this in an explicit way than a implicit one.
> 
> Should have explained it a bit better.

pgarcias01: Sorry @rennerocha I got it wrong, now i changed the start of the range

processing/data_collection/gazette/spiders/mg_itauna.py
```diff
@@ -17,7 +17,7 @@ def parse(self, response):
         last_page = int(
             response.xpath("//select[@id='select']/option[last()]/text()").get()
         )
-        for page in range(1, 1 + last_page):
+        for page in range(2, 1 + last_page):
             yield Request(page_url.format(page=page), callback=self.parse_editions_page)
```
giuliocc: Hey @pgarcias01 , right now the spider should get every page but the first one because the `parse` method only "connects" with the `parse_editions_page` method through the requests you are making. You should be able to solve it by adding this:
```suggestion
        for page in range(2, 1 + last_page):
            yield Request(page_url.format(page=page), callback=self.parse_editions_page)

		yield from self.parse_editions_page(response)
```
Since `parse_editions_page` is a generator, this way you can yield everything that is yielded by it "as if it was yielded by `parse` itself". In this case, you would be parsing the first page and yielding the item.

pgarcias01: Hi @giuliocc done! Thanks, again.

pgarcias01:
@giuliocc i can't change your request for review, can you do this?

processing/data_collection/gazette/spiders/mg_itauna.py
giuliocc: Hi @pgarcias01 I'm sorry I made an identation error there (github mobile). The `yield from ...` statement should be outside the `for` loop in this case. Could you fix it? ;/

pgarcias01: > Hi @pgarcias01 I'm sorry I made an identation error there (it should be correct by now). The `yield from ...` statement should be outside the `for` loop in this case. Could you fix it? ;/

@giuliocc done!

giuliocc (approved):
LGTM

rennerocha (approved):

rennerocha:
Thanks @pgarcias01 for the PR. The spider is working as expected!

PR #230, opened 2020-09-08
Add spider for Nova Serrana, MG

pgarcias01:
Add spider for Nova Serrana, MG

### DIFF
--- a/processing/data_collection/gazette/spiders/mg_novaserrana.py
+++ b/processing/data_collection/gazette/spiders/mg_novaserrana.py
@@ -1,0 +1,39 @@
+import datetime as dt
+from dateparser import parse
+from scrapy import Request
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class MgNovaSerranaSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3145208"
+    name = "mg_novaserrana"
+    allowed_domains = ["novaserrana.mg.gov.br"]
+    start_urls = ["https://www.novaserrana.mg.gov.br/portal/diario-oficial/"]
+
+    def parse(self, response):
+        page_url = response.urljoin("{page}/0/0/0/0/")
+        last_page = int(
+            response.xpath("//select[@id='select']/option[last()]/text()").get()
+        )
+        for page in range(2, 1 + last_page):
+            yield Request(page_url.format(page=page), callback=self.parse_editions_page)
+        yield from self.parse_editions_page(response)
+
+    def parse_editions_page(self, response):
+        diarios = response.css(".d_e_modelo_diario")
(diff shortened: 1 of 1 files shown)


giuliocc:
Hey! Great to see you contributing again :)

I noticed the spider looks pretty similar to the one you submitted recently (Itauna). Maybe there's a pattern here for multiple cities similar to those two?

pgarcias01:
> Hey! Great to see you contributing again :)
> 
> I noticed the spider looks pretty similar to the one you submitted recently (Itauna). Maybe there's a pattern here for multiple cities similar to those two?

Hi @giuliocc, we have other cities which are developed by the same company, being possible to use the same spyder. How do you suggest a way to add these cities?

giuliocc:
I think that creating a parent spider class with most of the base code and child spider classes for each city is a good practice in this case. If they don't differ much, only a few changes are necessary for each child. Most spiders from SC (Santa Catarina) use this, you could take a look to get some inspiration :)

rennerocha:
**question**: @pgarcias01 do you have a list of all cities that shares the same pattern?

**suggestion**: This spider is exactly equal to the one from Itauna-MG. Apparently there is no reason for code duplication, especially if we have more than two cities sharing the same pattern. Looking at the bottom of the pages it seems that they were developed by a company named Instar. You c ou

```
class BaseInstarSpider(BaseGazetteSpider):
    # Put your shared logic here
```

And then for each city you just create a subclass containing only the information required by that city:

```
class MgNovaSerranaSpider(BaseGazetteSpider):
    TERRITORY_ID = "3145208"
    name = "mg_novaserrana"
    allowed_domains = ["novaserrana.mg.gov.br"]
    start_urls = ["https://www.novaserrana.mg.gov.br/portal/diario-oficial/"]
```

rennerocha:
**thought**: Instar has a list of city clients [here](https://www.instar.com.br/desenvolvimento-de-sites/orgaos-publicos/13/portfolio.html). Maybe we could create in another issue a list of all these cities and verify which of them can use this shared code.

pgarcias01:
> **thought**: Instar has a list of city clients [here](https://www.instar.com.br/desenvolvimento-de-sites/orgaos-publicos/13/portfolio.html). Maybe we could create in another issue a list of all these cities and verify which of them can use this shared code.

Hi @rennerocha, I had already looked at this page, but not all of them use the diary page and there are other cities that are not there. I will close this PR and set up one with all that I found in the way they suggested.

rennerocha:
@pgarcias01 there is no need to close this PR.
We don't need to do all the cities at once. I think that if you just create a base spider for Instar pages, and just Nova  Serrana spider inheriting from it, is enough to start. Then the other cities can be new PRs that will reuse this base spider.

PR #231, opened 2020-09-09
Update Maceio-AL spider

rennerocha:
This PR updates Maceio-AL spider with improvements:
- Sort imports
- Replace MUNICIPALITY_ID (deprecated field) with TERRITORY_ID
- Replaced some insecure selectors
- Reorganiza pagination requests

### DIFF
--- a/processing/data_collection/gazette/spiders/al_maceio.py
+++ b/processing/data_collection/gazette/spiders/al_maceio.py
@@ -1,52 +1,50 @@
-from dateparser import parse
 from datetime import datetime
 
 import scrapy
-
+from dateparser import parse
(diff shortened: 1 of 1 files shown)


processing/data_collection/gazette/spiders/al_maceio.py
```diff
@@ -1,52 +1,50 @@
-from dateparser import parse
 from datetime import datetime
 
 import scrapy
-
+from dateparser import parse
 from gazette.items import Gazette
```
jvanz, line 5:
```suggestion 

from gazette.items import Gazette
```

jvanz (changes requested):

PR #233, opened 2020-09-17
Adding spider for Presidente Prudente/SP

GabrielTrettel:
This PR aims to merge[ Presidente Prudente (SP)](https://www.gdoe.com.br/presidenteprudente/) spider. But, it has three main problems that I need some help:

- I suppose my docker is not working properly, because after collecting the first 3 or 4 pages, it freezes. I waited for 1 hour and then killed it. This freezing behavior also occurs in the Santos spider, so I think that this is a local problem.

- On line 44: I am not yet familiar with the "official diaries", but it seems to me that the power is only the executive, but have no sure.

- In line 47: I don't know what an extra edition is, but I didn't see that term in the name of any file. Am I not looking at something?


processing/data_collection/gazette/spiders/sp_presidente_prudente.py
```diff
@@ -0,0 +1,53 @@
...
+from datetime import datetime
+from scrapy import Request
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class SpPresidentePrudenhteSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3541406"
+    GAZETTE_URL = "https://www.gdoe.com.br/presidenteprudente/1"
+    allowed_domains = ["https://www.gdoe.com.br/", "gdoe.com.br"]
```
jvanz, line 11: Take a look in the warning that I got running the spider:
```
2020-09-29 00:56:40 [py.warnings] WARNING: /usr/local/lib/python3.7/site-packages/scrapy/spidermiddlewares/offsite.py:61: URLWarning: allowed_domains accepts only domains, not URLs. Ignoring URL entry https://www.gdoe.com.br/ in allowed_domains.
  warnings.warn(message, URLWarning)
```

I have not check the scrapy documentation yet. But I **think** you can just remote the fist item in the list. Can you double check that? Thanks!

jvanz (changes requested):

jvanz:
>     * I suppose my docker is not working properly, because after collecting the first 3 or 4 pages, it freezes. I waited for 1 hour and then killed it. This freezing behavior also occurs in the Santos spider, so I think that this is a local problem.

Any error or warning message?

>     * On line 44: I am not yet familiar with the "official diaries", but it seems to me that the power is only the executive, but have no sure.

Yeah, I think you`re right. All the files seem to be from the executive power

>     * In line 47: I don't know what an extra edition is, but I didn't see that term in the name of any file. Am I not looking at something?

Usually, you can see in the web page when the gazettes is an extra edition. If that info is not present in the web page there is nothing you can do in the spider to identify it.

processing/data_collection/gazette/spiders/sp_presidente_prudente.py
GabrielTrettel: Yes, you are right. I guess this URL was there because of another method I tried before. Due to [docs](https://docs.scrapy.org/en/latest/topics/spiders.html#scrapy.spiders.Spider.allowed_domains) I will just remove the string in my next commit.

processing/data_collection/gazette/spiders/sp_presidente_prudente.py
```diff
@@ -8,7 +8,7 @@
 class SpPresidentePrudenhteSpider(BaseGazetteSpider):
```
rennerocha, line 8: **nitpick** There is a type on class name. It should be `SpPresidentePrudenteSpider`.

GabrielTrettel: I will send a commit to fix this typo.

GabrielTrettel: Not so nitpick, but I should **definitely** have created my own branch for this PR. Noob here.

processing/data_collection/gazette/spiders/sp_presidente_prudente.py
```diff
@@ -0,0 +1,53 @@
+import re
+from datetime import datetime
+from scrapy import Request
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
```
jvanz, line 5:
```suggestion
import re
from datetime import datetime

from scrapy import Request

from gazette.items import Gazette
from gazette.spiders.base import BaseGazetteSpider
```

jvanz (approved):

Issue #221, opened 2020-08-21
Finish old spider PRs


jvanz:
There are a bunch of old  PRs adding spiders with no recent activity. It would be great if somebody review, fix and open a PR for them.

chamun:
I would like to help with this issue. 

I thought I could start with https://github.com/okfn-brasil/querido-diario/pull/70 which seems stale for more than 2 years. Should I just open a new PR once I'm done?

jvanz:
> I would like to help with this issue.
> 
> I thought I could start with #70 which seems stale for more than 2 years. Should I just open a new PR once I'm done?

Yeah, that's right! ;)

Lucs1590:
Is there any spider still pending?

Lucs1590:
@jvanz , I think the correct label would be hacktoberfest-accepted, what do you think?
I based on the following link: https://hacktoberfest.digitalocean.com/hacktoberfest-update

rennerocha:
@Lucs1590 this is for PR, not for issues. This label indicates that this is a good issue for anyone that wants to participate on Hacktoberfest.

Lucs1590:
Very nice, you right @rennerocha !
Thank you for response.

begnini:
I created 3 spiders #522 #521 and #520, fixing #22, #26 and #75. I think we can close these old PRs.

rennerocha:
This is issue is too generic. We can deal with the open PRs directly.

PR #70, opened 2018-06-01
Add spider for Campos dos Goytacazes/RJ

dannnylo:
In this PR we add the spider to collect data from [Campos dos Goytacazes/RJ gazette](https://www.campos.rj.gov.br/diario-oficial.php)

PR #235, opened 2020-09-17
Add spider for Campos dos Goytacazes/RJ

chamun:
This PR continues the work started in #70 as part of #221.

It addresses the change requests from #70, merges main into the branch and fixes the code to download all the gazettes since 2015.


chamun:
The failing CI error is unrelated to this PR. I have opened https://github.com/okfn-brasil/querido-diario/pull/236 to address this problem.

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
```diff
@@ -0,0 +1,67 @@
...
+class RjCampoGoytacazesSpider(BaseGazetteSpider):
+    GAZETTE_ELEMENT_CSS = "ul.ul-licitacoes li"
+    TERRITORY_ID = "3301009"
+
+    allowed_domains = ["www.campos.rj.gov.br"]
+    name = "rj_campos_goytacazes"
+    start_urls = [
+        "https://www.campos.rj.gov.br/diario-oficial.php?PGpagina=1&PGporPagina=15"
+    ]
+    LIMIT_DATE = dateparser.parse("2015-01-01").date()
```
rennerocha, line 20: **suggestion**: If you are creating a `datetime.date` object, there is no need to use an intermediary like `dateparser`.

[Built-in date](https://docs.python.org/3/library/datetime.html#datetime.date) from Python is enough:

`LIMIT_DATE = datetime.date(2015, 1, 1)`

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
```diff
@@ -0,0 +1,67 @@
...
+
+from datetime import datetime
+from scrapy import Request
+from scrapy.exceptions import CloseSpider
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class RjCampoGoytacazesSpider(BaseGazetteSpider):
+    GAZETTE_ELEMENT_CSS = "ul.ul-licitacoes li"
```
rennerocha, line 12: **suggestion**: Considering that we only use this selector once, and we have many other selectors inside `parse` method, in my opinion having this here is not helpful. If we need to change it because the layout changed, it is very likely that we will have to update the entire spider anyway.

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
```diff
@@ -0,0 +1,67 @@
...
+    def parse(self, response):
+        """
+        @url https://www.campos.rj.gov.br/diario-oficial.php?PGpagina=1&PGporPagina=15
+        @returns requests 1
+        @returns items 15 15
+        @scrapes date file_urls is_extra_edition municipality_id power scraped_at
+        """
+
+        for element in response.css(self.GAZETTE_ELEMENT_CSS):
+            gazette_text = element.css("h4::text").extract_first() or ""
```
rennerocha, line 31: **suggestion** Scrapy suggest the use of `.get` instead of (deprecated) `extract_first`. 

**suggestion** `get` (and `extract_first`) accepts a `default` [argument](https://parsel.readthedocs.io/en/latest/parsel.html#parsel.selector.SelectorList.get), so you don't need to add this `or ""`

This is equivalent to:
`gazette_text = element.css("h4::text").get("")`

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
```diff
@@ -0,0 +1,67 @@
...
+
+            date_re = re.search(r"(\d{2} de (.*) de \d{4})", gazette_text)
```
rennerocha, line 33: **suggestion** If you are using `gazette_text` just to apply a regex over it, you can use `re` method:

This will return a tuple with 3 items containing the information you want:
`date = element.css("h4::text").re('r"((\d{2}) de (.*) de (\d{4})')`
It will return `None` if it didn't match, so you can check it as you did it next.

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
```diff
@@ -0,0 +1,67 @@
...
+            if not date_re:
+                continue
+
+            # The extra edition for August 28th, 2018 has a typo in the month name.
+            date = date_re.group(0).replace("Agosoto", "Agosto")
```
rennerocha, line 38: **praise** Good catch :-) This kind of error is difficult to find sometimes.

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
```diff
@@ -0,0 +1,67 @@
...
+            date = dateparser.parse(date, languages=["pt"]).date()
+
+            if date < self.LIMIT_DATE:
```
rennerocha, line 41: **question** Why you are limiting the date? If we have older gazettes, we should get it as well.

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
```diff
@@ -0,0 +1,67 @@
...
+                raise CloseSpider("Went further than 2015")
+
+            url = element.css("a::attr(href)").extract_first().strip()
+            # From November 17th, 2017 and backwards the path to the gazette PDF
```
rennerocha, line 45: **suggestion** If the URL is relative, but the netloc is still the same, you can use `response.urljoin` to make the absolute URL:
`url = response.urljoin(element.css("a::attr(href)").get())`

rennerocha (changes requested):

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
chamun: Because [CONTRIBUTING.md](https://github.com/okfn-brasil/querido-diario/blob/main/CONTRIBUTING.md) states that the projec'ts goal is  to download gazettes since 2015:

> At this moment, the project aims for the 100 top Brazilian municipalities by population, collecting all the gazettes since 2015 

Campos Goytacazes City Hall's website provides gazettes up until 2010.

Should I remove this limit?

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
chamun: Nice!

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
chamun: `gazette_text` is also used to set the value of `is_extra_edition`.

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
chamun: I'll address this once we settle on https://github.com/okfn-brasil/querido-diario/pull/235#discussion_r491440601.

chamun:
@rennerocha this PR is ready for another look.

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
```diff
@@ -0,0 +1,66 @@
+import re
+import dateparser
+
+from datetime import datetime
+from scrapy import Request
+from scrapy.exceptions import CloseSpider
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
```
jvanz, line 8:
```suggestion
from datetime import datetime
import re

import dateparser
from scrapy import Request
from scrapy.exceptions import CloseSpider

from gazette.items import Gazette
from gazette.spiders.base import BaseGazetteSpider
```

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
jvanz: I think you forget to update this line. Because I cannot see the change in the highlighted comment. Am I wrong?

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
jvanz: Yes. Please, get all the files available. I think what that line was trying to say is that the spiders in the time of that writing was crawling files since 2015.

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
jvanz: :+1:

jvanz (changes requested):

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
chamun: I did not update this line because I was waiting on a reply on my question in https://github.com/okfn-brasil/querido-diario/pull/235#discussion_r491440601. I'll update it now!

processing/data_collection/gazette/spiders/rj_campos_goytacazes.py
chamun: @jvanz OK, I'll work on that!

chamun:
@jvanz @rennerocha All the change requests have been addressed.

jvanz (approved):
LGTM. 

For the maintainer who merge the PR. Please, use "squash and merge option" in the web interface.

Issue #247, opened 2020-09-30
Reduce number of requests made in partial scrapes


rennerocha:
In order to check the execution time of the spiders, I deployed the project on [Scrapy Cloud](https://docs.scrapy.org/en/latest/topics/deploy.html#deploying-to-scrapy-cloud) and selected manually a few spiders.

During my tests I did the follow:
- Disabled convert PDF to text step (as Scrapy Cloud doesn't allows us to configure pdf2txt in a free tier);
- Took advantage of [GazetteDateFilteringPipeline](https://github.com/okfn-brasil/querido-diario/blob/2c3c722969b3f291214d30a43a07cdf1605c8052/processing/data_collection/gazette/pipelines.py#L17) and set `start_date` to avoid downloading more gazettes than we want.

Results were not good. Even if we restrict the date of the gazettes that we want to download, the number of unnecessary requests made the spider take longer than accepted.

Considering that each city site is different, I couldn't think a general solution that may be applied to all spiders to reduce the number of requests other than changing every spider and including some internal logic that will filter the requests based on how data is structured there.

spider | elapsed (min) | date
------------ | ------------- | ---
sp_jundiai | 49 | all
sp_jundiai | 13 | last day
ba_salvador | 33 | last day
to_palmas | 42 | current month

Is there any other option to improve this?

vitorbaptista:
Do we know where this time is spent on? Are there any unecessary requests being made? Maybe a quick and dirty way to find potential spiders to optimize is looking for the ratio between number of requests and number of items scraped. To scrape a single gazette, how many requests are needed? If a scraper needs a high number of requests per gazette it might mean that it's doing unecessary requests.

rennerocha:
> Do we know where this time is spent on? Are there any unecessary requests being made? Maybe a quick and dirty way to find potential spiders to optimize is looking for the ratio between number of requests and number of items scraped. To scrape a single gazette, how many requests are needed? If a scraper needs a high number of requests per gazette it might mean that it's doing unnecessary requests.

We **definitely** send unnecessary requests. I would like to fix it without changing each spider, but I don't think it will be possible given that each city organize its gazettes in a different way.

The pattern that we use to develop all spiders crawl the entire site for all items and yield them. We avoid downloading the gazettes based using an Item Pipeline that drops items older than a specified data (`start_date` argument). It is fine when doing a full scrape but is costly for partial scrapes.

[...]

PR #251, opened 2020-10-01
Natal-RN - Limit requests according to start date argument provided

rennerocha:
Based on the discussions on #247 , I did a change in Natal-RN spider and introduce a smarter procedure stop sending more requests than we need if we want to get only gazettes newer than an specific date.

Comparing execution time and number of requests we can see a huge difference between the code in `main` and this PR code:

natal_spiders

For sites where we can decide to send requests providing search arguments (like year and month), this approach should be enough to make them faster for incremental scrapes.

### DIFF
--- a/processing/data_collection/gazette/spiders/rn_natal.py
+++ b/processing/data_collection/gazette/spiders/rn_natal.py
@@ -1,6 +1,7 @@
-from datetime import datetime
+from datetime import date, datetime
 
 import dateparser
+from dateutil.rrule import rrule, MONTHLY
 from scrapy import FormRequest
 
 from gazette.items import Gazette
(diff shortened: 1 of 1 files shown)


processing/data_collection/gazette/spiders/rn_natal.py
```diff
@@ -11,15 +12,23 @@ class RnNatalSpider(BaseGazetteSpider):
 
     name = "rn_natal"
     allowed_domains = ["www.natal.rn.gov.br"]
+    start_date = date(2003, 1, 1)
```
ejulio, line 15: I guess it should be `"2003-1-1"` so you can parse the input string in `start_requests`, otherwise you won't be able to set it through the CLI

ejulio (changes requested):

processing/data_collection/gazette/spiders/rn_natal.py
ejulio: My bad... :sweat:

ejulio (approved):

ejulio (approved):

PR #253, opened 2020-10-01
Add `DefaultValuesPipeline`

tcurvelo:
For the sake of DRY, I added an Item Pipeline for setting default values for `territory_id` and `scraped_at`. 
Any comments are welcome.

### DIFF
--- a/processing/data_collection/gazette/pipelines.py
+++ b/processing/data_collection/gazette/pipelines.py
@@ -1,9 +1,9 @@
-from pathlib import Path
-import hashlib
-import magic
+import datetime as dt
 import os
 import subprocess
+from pathlib import Path
 
+import magic
 from itemadapter import ItemAdapter
 from scrapy.exceptions import DropItem
 from scrapy.http import Request
@@ -19,6 +19,21 @@ class GazetteDateFilteringPipeline:
         if hasattr(spider, "start_date"):
             if spider.start_date > item.get("date"):
                 raise DropItem("Droping all items before {}".format(spider.start_date))
+        return item
+
+
+class DefaultValuesPipeline:
+    """ Add defaults values field, if not already set in the item """
+
+    default_field_values = {
+        "territory_id": lambda item, spider: getattr(spider, "TERRITORY_ID"),
+        "scraped_at": lambda item, spider: dt.datetime.utcnow(),
+    }
+
+    def process_item(self, item, spider):
+        for field in self.default_values:
+            if field not in item:
+                item[field] = self.default_field_values(field)(item, spider)
         return item
 
 
--- a/processing/data_collection/gazette/pipelines.py
+++ b/processing/data_collection/gazette/pipelines.py
@@ -26,14 +26,14 @@ class DefaultValuesPipeline:
     """ Add defaults values field, if not already set in the item """
 
     default_field_values = {
-        "territory_id": lambda item, spider: getattr(spider, "TERRITORY_ID"),
-        "scraped_at": lambda item, spider: dt.datetime.utcnow(),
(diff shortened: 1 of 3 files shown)


processing/data_collection/gazette/pipelines.py
```diff
@@ -22,6 +22,21 @@ def process_item(self, item, spider):
...
+class DefaultValuesPipeline:
+    """ Add defaults values field, if not already set in the item """
+
+    default_field_values = {
+        "territory_id": lambda item, spider: getattr(spider, "TERRITORY_ID"),
+        "scraped_at": lambda item, spider: dt.datetime.utcnow(),
+    }
+
+    def process_item(self, item, spider):
+        for field in self.default_values:
```
rennerocha, line 34: **issue**: `self.default_values` is not defined. Isn't it `self.default_field_values`?

processing/data_collection/gazette/pipelines.py
```diff
@@ -22,6 +22,21 @@ def process_item(self, item, spider):
...
+            if field not in item:
+                item[field] = self.default_field_values(field)(item, spider)
```
rennerocha, line 36: **issue**: `self.default_field_values` is a `dict`. So it is not callable. I believe that `self.default_field_values[field](item, spider)` will solve that.

**question**: why are you passing `item` to the functions? It is not used by any of them.

processing/data_collection/gazette/pipelines.py
```diff
@@ -22,6 +22,21 @@ def process_item(self, item, spider):
         return item
 
 
+class DefaultValuesPipeline:
+    """ Add defaults values field, if not already set in the item """
+
+    default_field_values = {
+        "territory_id": lambda item, spider: getattr(spider, "TERRITORY_ID"),
```
rennerocha, line 29: **suggestion**: If a spider doesn't have a `TERRITORY_ID` set, it will raise an `AttributeError`. You could add a default value if this happens: `getattr(spider, "TERRITORY_ID", "UNKNOWN")`.

processing/data_collection/gazette/pipelines.py
```diff
@@ -22,6 +22,21 @@ def process_item(self, item, spider):
         return item
 
 
+class DefaultValuesPipeline:
```
rennerocha, line 25: **praise**: I liked this pipeline very much. It doesn't make sense to add extra lines to the spider with values that can be included automatically.

rennerocha (changes requested):
@tcurvelo thanks for your PR. It will be very helpful. I found some issues that prevents this PR to be merged as-it-is now. Please take a look at my comments and please let me know if you have some questions about them

processing/data_collection/gazette/pipelines.py
ejulio: I think this is intended, right?
Shouldn't all spiders should define a `TERRITORY_ID`? And in case it is missing, it should fail.
Maybe, this is not the right place to raise this error, but if the field is required, it should be checked somewhere to avoid yielding items without this info

processing/data_collection/gazette/pipelines.py
tcurvelo: You're right about both issues. They were due a (bad) last minute refactor. My fault. :facepalm: 
Regarding the `item`, I added it "just in case" someone wants to add another default value that used it. But, yes it doesn't make sense right now. Let me remove it for the sake of YAGNI :)

processing/data_collection/gazette/pipelines.py
tcurvelo: Yeah, good points. Both of you :)
Let me add a check for that at `base.py`

jvanz (approved):
:heart:

rennerocha (approved):

PR #217, opened 2020-08-20
Add new spider for Macapa-AP

rennerocha:
This PR solves https://github.com/okfn-brasil/querido-diario/issues/180.

Macapa page only provide gazettes from 2020.
Some of the gazettes are images, so maybe their content will not be retrievable.

Issue #180, opened 2020-06-20
Macapá spider


jvanz:
Add an spider for Macapá, capital of Amapá

phrfpeixoto:
I'll be working on this one. Or at least I'll try.

rennerocha:
@phrfpeixoto there is already a PR open of a Macapá spider almost ready for merge.

phrfpeixoto:
Oh, sure. My bad, I didn't see that reference up there.

PR #254, opened 2020-10-02
Adding a spider for AP Macapá

phrfpeixoto:
This is my proposed solution for #180.

I know @rennerocha commented on that issue, PR #217 is already under review for this spider, but by the time I saw that comment, the code was already written, so I figured I should submit it anyway. At least I can get some comments and learn something.


phrfpeixoto:
I opted to perform one request per day, since lots of the entries on the website do not carry a date specification anywhere on the HTML body.

I also noticed that the URL path (which includes a date) is not always a good representation for the actual date for that gazette.

This will result in much more requests, but it's the way I found to assure that dates will be consistent, and no results will be skipped due to the lack of a date.

phrfpeixoto:
Sorry! I unwillingly closed this trying to rebase my branch. Reopened.

processing/data_collection/gazette/spiders/ap_macapa.py
```diff
@@ -0,0 +1,79 @@
...
+
+        target_date = self.start_date
+        data = {
+            "s": "",
+            "post_type": "official_diaries",
+            "search": "official_diaries",
+            "official_diary_number": "",
+        }
+        while target_date <= self.end_date:
+            formatted_date = target_date.strftime(date_format)
```
ejulio, line 46: It seems that `date_format` is only used here, so I'd use the string here and not define a new variable.
It would be easier to read, I think

ejulio (approved):

processing/data_collection/gazette/spiders/ap_macapa.py
```diff
@@ -0,0 +1,79 @@
...
+    name = "ap_macapa"
+    allowed_domains = ["macapa.ap.gov.br"]
+    start_date = None
+
+    TERRITORY_ID = "1600303"
+
+    def __init__(self, start_date=None, end_date=None, *args, **kwargs):
+        if start_date is None:
+            # Start from yesterday
+            start_date = dt.date.today() + dt.timedelta(days=-1)
```
jvanz, line 21: I think this way is more readable: 

```suggestion
            start_date = dt.date.today() - dt.timedelta(days=1)
```

processing/data_collection/gazette/spiders/ap_macapa.py
```diff
@@ -0,0 +1,79 @@
...
+            start_date = dateparser.parse(start_date)
+
+        if end_date is None:
+            # End today
+            end_date = dt.date.today()
+        else:
+            end_date = dateparser.parse(end_date)
+
+        self.start_date = start_date
+        self.end_date = end_date
```
jvanz, line 32: I'm considering the idea of moving all this logic of start and end date to the base class. Thus, we can try to add the same idea for other spiders, when possible. Please, update the PR.

jvanz (changes requested):

processing/data_collection/gazette/spiders/ap_macapa.py
phrfpeixoto: Done.

processing/data_collection/gazette/spiders/ap_macapa.py
phrfpeixoto: Done. The changes to the base class are proposed on #265.

processing/data_collection/gazette/spiders/ap_macapa.py
phrfpeixoto: Removed that. No longer applicable.

processing/data_collection/gazette/spiders/ap_macapa.py
```diff
@@ -0,0 +1,73 @@
...
+class ApMacapaSpider(BaseGazetteSpider):
+
+    name = "ap_macapa"
+    allowed_domains = ["macapa.ap.gov.br"]
+    start_date = None
+
+    TERRITORY_ID = "1600303"
+
+    def __init__(self, start_date=None, end_date=None, *args, **kwargs):
+        self.start_date = dt.date.today().replace(day=1)
```
jvanz, line 18: Please, set the `start_date` to the first date with gazettes available. Other spiders already do that as well. Thanks

jvanz (changes requested):

processing/data_collection/gazette/spiders/ap_macapa.py
phrfpeixoto: I though the idea was having incremental crawls, but ok, I'll do it.

processing/data_collection/gazette/spiders/ap_macapa.py
```diff
@@ -0,0 +1,72 @@
...
+                {
+                    "official_diary_initial_date": formatted_date,
+                    "official_diary_final_date": formatted_date,
+                }
+            )
+
+            yield scrapy.FormRequest(
+                url=base_url, formdata=data, method="GET", meta={"date": target_date}
+            )
+            target_date = target_date + dt.timedelta(days=1)
```
jvanz, line 50: If you can define the date range, why return one request per day? Why not just one single request with the date range that you want? Image that the user wants de files from a month. In this version, you will send request 30~31 requests instead of 1.

phrfpeixoto: That is explained [here](https://github.com/okfn-brasil/querido-diario/pull/254#issuecomment-702490394), but in more detail:

That's an issue with the site: There's no reliable way to identify the correct date of the gazette by looking at the HTML code:
While most of the gazettes do have a date written on the posting, some do not:
Screen Shot 2020-10-08 at 09 55 16
And the code for these entries look like this:
```html
<div class="panel panel-default">
  <div class="panel-heading"><h4>Diário N° 3245</h4></div>
  <div class="panel-body">
    <a href="https://macapa.ap.gov.br/portal/wp-content/uploads/2019/02/DOM_3245_2018.pdf" target="_blank">
        <i class="fa fa-file-pdf-o" aria-hidden="true"></i><h4>DIÁRIO OFICIAL - N° 3245/ 2018</h4>
    </a>
  </div>
</div>
```
We can clearly see a date indication on the URL, but this is a clearly Wordpress website, and the date within the path reflects the date on which the file has been uploaded to the blog, not the date on which the gazette was issued. I did look around and found some instances where these date do not match.

If you can think of a better way to find a reliable date, please let me know.

jvanz: I understand now. Thanks for the explanation and the links.

jvanz (approved):

Issue #256, opened 2020-10-02
Avoid downloading PDFs while testing


phrfpeixoto:
The current code for the the `ExtractTextPipeline` blindly accesses the first element on the `item['files']` list, on line 37:

https://github.com/okfn-brasil/querido-diario/blob/a241eba2d41c2dca9bfa70565f954e6c2dfcc3f3/processing/data_collection/gazette/pipelines.py#L25-L41

This prevents the developer from crawling gazettes while avoiding downloading large PDF files (which on some sites can be slow), which is very useful while writing the spider code and testing agains the live website.

phrfpeixoto:
Similar bug in `SQLDatabasePipeline`

jvanz:
Text extraction has been removed form the pipeline. Closing this issue

PR #257, opened 2020-10-02
Skipping ExtractTextPipeline if the item has no files

phrfpeixoto:
This is a simple fix for #256

### DIFF
--- a/processing/data_collection/gazette/pipelines.py
+++ b/processing/data_collection/gazette/pipelines.py
@@ -34,6 +34,9 @@ class ExtractTextPipeline:
         if not extract_text_from_file:
             return item
 
+        if len(item["files"]) == 0:
+            return item
+
         if self.is_doc(item["files"][0]["path"]):
             item["source_text"] = self.doc_source_text(item)
         elif self.is_pdf(item["files"][0]["path"]):


processing/data_collection/gazette/pipelines.py
```diff
@@ -34,6 +34,9 @@ def process_item(self, item, spider):
         if not extract_text_from_file:
             return item
 
+        if len(item["files"]) == 0:
```
ejulio, line 37: I guess `not item["files"]` reads better, unless you want to check specifically for an empty list

ejulio (approved):

jvanz:
Thanks for your PR @phrfpeixoto ! But I`ll not accept it. We are in the process of moving off the data manipulation code from this repository. So, this code will be obsolete within days. Sorry... =/

If you interesting in check the future commits on that. Keep eyes on this repo: https://github.com/okfn-brasil/querido-diario-data-processing

phrfpeixoto:
Added another check for `SQLDatabasePipeline`, which also has a bug. And this time, no way to work around it.
@giuliocc @rennerocha @ejulio 
Can we re-open this PR?

processing/data_collection/gazette/pipelines.py
phrfpeixoto: Fixed.

ejulio:
@phrfpeixoto , no new commits here, are you sure you pushed it?
Also, if the change is in another pipeline, why not making it a new PR?

phrfpeixoto:
Will do

giuliocc:
@phrfpeixoto for this pipeline, I think we should wait for #310. If there's a bug in the SQL pipeline I suggest opening a new PR for solving it.

Issue #258, opened 2020-10-02
Mossoró Spider


phrfpeixoto:
Add a spider for Mossoró/RN

rennerocha:
Closed by https://github.com/okfn-brasil/querido-diario/pull/259

PR #85, opened 2018-06-07
Add spider for Mossoró/RN

agessner:
In this PR we add the spider to collect data from [Mossoró/RN gazettes](https://www.prefeiturademossoro.com.br/jom/alljoms.php).

We have found that the gazette's dates in the website are not  trustworthy:
- We have a gazette with a date that doesn't exist, `June 31st, 2017`. But inside the gazette the date is `May 31st, 2017`. Right now, it breaks the code when it tries to parse the date.
How should we proceed? We thought about reading the gazette content and get the date from there. But right now we couldn't find a way to do that. Any thoughts?

- We have dates in future, like `December 17th, 2018`.  But inside the gazette the date is `January 17th, 2018`. It is similar to the previous problem.

- We have gazettes that are duplicated, like 2018-01-12, 2017-12-20, 2017-10-31, etc...
About that, we thought that we could check if a given gazette's date has the same date from another gazette that was already processed, if so we ignore it. Is it a good approach?

Co-authored-by: @LucasSedrez and @rhogeranacleto

PR #259, opened 2020-10-02
Adding a spider for RN - Mossoró

phrfpeixoto:
Spider proposal for #258

### DIFF
--- a/processing/data_collection/gazette/spiders/rn_mossoro.py
+++ b/processing/data_collection/gazette/spiders/rn_mossoro.py
@@ -1,0 +1,65 @@
+# coding=utf-8
+import re
+from datetime import datetime, date
+
+import dateutil
+import scrapy
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class RnMossoroSpider(BaseGazetteSpider):
+
+    name = "rn_mossoro"
+    allowed_domains = ["jom.prefeiturademossoro.com.br"]
+
+    TERRITORY_ID = "2408003"
+
+    def __init__(self, start_date=None, end_date=None, *args, **kwargs):
+        self.start_date = date(2008, 1, 1)
+        self.end_date = date.today()
+
+        super(RnMossoroSpider, self).__init__(start_date, end_date, *args, **kwargs)
+
+        self.logger.debug(
+            "Start date is {date}".format(date=self.start_date.isoformat())
+        )
+        self.logger.debug("End date is {date}".format(date=self.end_date.isoformat()))
+
+    def start_requests(self):
+
+        for year in range(self.start_date.year, self.end_date.year + 1):
+            for month in range(self.start_date.month, self.end_date.month + 1):
+                base_url = f"http://jom.prefeiturademossoro.com.br/{year}/{month}/"
+                yield scrapy.Request(base_url)
+
+    def parse(self, response):
+        for entry in response.css("article.post.category-jom"):
+
+            url = entry.css("a:first-of-type::attr(href)").get()
+            datetime_meta = entry.css("time.published::attr(datetime)").get()
+            entry_date = dateutil.parser.isoparse(datetime_meta).date()
+
+            yield scrapy.Request(
+                url, meta={"date": entry_date}, callback=self.parse_gazette
+            )
+
+        next_pages_links = response.css("a.page-numbers::attr(href)").getall()
+        for link in next_pages_links:
+            yield scrapy.Request(link)
+
+    def parse_gazette(self, response):
+        gazette_date = response.meta["date"]
+        file_url = response.css(".wp-block-file__button::attr(href)").get()
+        edition_regex = re.compile("JOM n\.º ([a-z0-9]+)$", re.IGNORECASE)
+        edition = response.css("h1.entry-title::text").re_first(edition_regex)
+        yield Gazette(
+            date=gazette_date,
+            edition_number=edition,
+            file_urls=[file_url],
+            is_extra_edition=False,
+            territory_id=self.TERRITORY_ID,
+            power="executive_legislative",
+            scraped_at=datetime.utcnow(),
+        )
--- a/CITIES.md
+++ b/CITIES.md
@@ -99,7 +99,7 @@ The municipality id (IBGE code) can be found on [Wikipedia](https://pt.wikipedia
 | 89 | Petrópolis | | | |
 | 90 | Camaçari | | | |
 | 91 | Santarém | | | |
-| 92 | Mossoró | | | |
+| 92 | Mossoró | :white_check_mark: | | [PR](https://github.com/okfn-brasil/querido-diario/pull/259) |
 | 93 | Suzano | | | |
 | 94 | Palmas | :white_check_mark: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/1) |
 | 95 | Governador Valadares | :soon: | | [PR](https://github.com/okfn-brasil/diario-oficial/pull/19) |


processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,50 @@
...
+
+            yield scrapy.Request(url, meta={"date": date}, callback=self.parse_gazette)
+
+        page = int(response.css(".pt-cv-pagination::attr(data-currentpage)").get())
+        total_pages = int(
+            response.css(".pt-cv-pagination::attr(data-totalpages)").get()
+        )
+
+        if page == 1:
+            for i in range(2, total_pages + 1):
```
ejulio, line 33: Given the conversation here https://github.com/okfn-brasil/querido-diario/issues/247
We should avoid yielding all pages upfront and walk them one each time, or something like that, to avoid making unnecessary requests

phrfpeixoto: Refactored. On my tests now it issues about 3 to 4 additional requests when compared to the amount of scraped items.

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class RnMossoroSpider(BaseGazetteSpider):
+
+    name = "rn_mossoro"
+    allowed_domains = ["jom.prefeiturademossoro.com.br"]
+
+    TERRITORY_ID = "2408003"
+
+    def __init__(self, start_date=None, end_date=None):
```
ejulio, line 17: I think you need to get `*args` and `**kwargs` here.
In case we set `-a spider_arg=123` in the CLI, they are set to `__init__()` as `**kwargs` and set as class attributes by the base spider.

Also, I'd expect the base class to set `start_date` and `end_date` properly, and only set them here in case we had to overwrite the value

phrfpeixoto: Fixed.

ejulio (approved):

processing/data_collection/gazette/settings.py
```diff
@@ -8,4 +8,4 @@
     "gazette.pipelines.ExtractTextPipeline": 200,
 }
 FILES_STORE = "/mnt/data/"
-QUERIDODIARIO_EXTRACT_TEXT_FROM_FILE = True
\ No newline at end of file
+QUERIDODIARIO_EXTRACT_TEXT_FROM_FILE = True
```
alexandrevicenzi, line 11: this same change is in another of your PRs, it will cause conflicts once that one is merged

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,63 @@
...
+
+class RnMossoroSpider(BaseGazetteSpider):
+
+    name = "rn_mossoro"
+    allowed_domains = ["jom.prefeiturademossoro.com.br"]
+
+    TERRITORY_ID = "2408003"
+
+    def __init__(self, start_date=None, end_date=None, *args, **kwargs):
+        self.start_date = date.today().replace(day=1)
```
alexandrevicenzi, line 18: would be better to set this value only if none right? and also, only after the super statement I think

alexandrevicenzi (changes requested):

processing/data_collection/gazette/settings.py
phrfpeixoto: We decided to rebase this once the first one gets merged.

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: That's the opposite, actually. The super class only touches `self.start_date` if the spider args are defined, so the idea here is:
1- Set default values
2- Override the default value if the user defined something valid.

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,63 @@
...
+        self.end_date = date.today()
+
+        super(RnMossoroSpider, self).__init__(start_date, end_date, *args, **kwargs)
+
+        self.logger.debug(
+            "Start date is {date}".format(date=self.start_date.isoformat())
+        )
+        self.logger.debug("End date is {date}".format(date=self.end_date.isoformat()))
```
giuliocc, line 26: The `end_date` attribute doesn't fit the current project pattern (although I like it and encourage discussion about it). Having a particular spider with a different set of parameters should only be done if absolutely necessary in my opinion.

Also, I don't think the logs add much. So right now I'm thinking this could become cleaner if the `__init__` method was removed. What do you think?

`start_date` as a class attribute and `end_date` as a variable in `start_requests` or something like that?

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def start_requests(self):
+
+        for year in range(self.start_date.year, self.end_date.year + 1):
+            for month in range(self.start_date.month, self.end_date.month + 1):
```
giuliocc, line 31: `dateutil.rrule` with a `MONTHLY` frequency could be handy. Just a suggestion if you don't already know it and finds it useful :)

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,63 @@
...
+                base_url = f"http://jom.prefeiturademossoro.com.br/{year}/{month}/"
+                yield scrapy.Request(base_url)
+
+    def parse(self, response):
+        for entry in response.css("article.post.category-jom"):
```
giuliocc, line 36: Using the classes in the selector doesn't seem necessary at a first glance. What do you think?

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,63 @@
...
+
+            url = entry.css("a:first-of-type::attr(href)").get()
+            datetime_meta = entry.css(
+                "div.post-data-text time.published::attr(datetime)"
```
giuliocc, line 40: `div.post-data-text` also doesn't seem necessary at a first glance and we could simplify this selector. What do you think?

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,63 @@
...
+            ).get()
+            entry_date = dateutil.parser.isoparse(datetime_meta).date()
+
+            yield scrapy.Request(
+                url, meta={"date": entry_date}, callback=self.parse_gazette
+            )
+
+        page_links = response.css("a.page-numbers::attr(href)").getall()
+        for link in page_links:
+            yield scrapy.Request(link)
```
giuliocc, line 50: Could we change `page_links` to `next_pages_links` or something like that to make it clearer that this is about pagination? I'm currently processing this as something like "every link in this page" hehehehe

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def parse_gazette(self, response):
+        date = response.meta["date"]
+        file_url = response.css(".wp-block-file__button::attr(href)").get()
```
giuliocc, line 54: Instead of looking for classes, what do you think about an xpath that looks for the text content of the element?
```suggestion
        file_url = response.xpath("//a[contains(./text(), 'Baixar')]/@href").get()
```

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,63 @@
...
+
+        yield Gazette(
+            date=date,
+            file_urls=[file_url],
+            is_extra_edition=False,  # This information is missing
```
giuliocc, line 59: Could you clarify what you mean by missing?

giuliocc (changes requested):
Hey @phrfpeixoto , this looks great! Since the spider is working like a charm, just some minor adjustments are needed. And I also gave you some suggestions. Let me know what you think :)

vitorbaptista:
I didn't see this PR, and ended up updating #85, merging with the current main branch and fixing the code to work with the current Mossoró website. The new PR is #296. Feel free to close that, if you prefer this PR instead.

Sorry for the mistake.

processing/data_collection/gazette/spiders/rn_mossoro.py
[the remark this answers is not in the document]
phrfpeixoto: It's been used on 3 or 4 other spiders I've submitted, and other reviewers seem to approve it.

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: It's been used on 3 or 4 other spiders I've submitted, and other reviewers seem to approve it.
Also, the logs are for debugging, they won't show up on "production" logs

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: I've used it before, but I find that this here is more readable. Far more people can read a `range()` call

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: Indeed. All the `<article />` tags here are gazettes, but IMO the more specific, the better.
They they add other "entities" to the page, that are also rendered  `<article />`, this will still work.

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: Again correct, just being more explicit and specific.

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: I find the classes to look more reliable. "Baixar" can be switched for an icon, or "Download", or "Obter", while the class should remain the same unless they redesign.

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: The information is no available anywhere on the website.

processing/data_collection/gazette/spiders/rn_mossoro.py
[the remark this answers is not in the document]
phrfpeixoto: Done.

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: Done.

phrfpeixoto:
@giuliocc Thanks for the review. I've commented and resolved the ones that I didn't agree, and left some that I'm open to change. Let me know your thoughts.

processing/data_collection/gazette/spiders/rn_mossoro.py
giuliocc: I just felt like it would make it more readable without losing important functionality. If you see advantage in it, there's no problem.

And for the `end_date` argument, nothing about this would block you PR. As I said, I like the idea, I just think it would be a good practice to discuss it before introducing it. Then we could develop things to incorporate it better like having an attribute in the base class, etc. But no worries about it.

processing/data_collection/gazette/spiders/rn_mossoro.py
giuliocc: As a disclaimer, I think that this discussion isn't too relevant. So I'll be trying to keep it short and this would never block your PR. I'll just wait for your closing thoughts on this.

Generic selectors are not always good for the reason you gave. But, the more specific is not always the better since you introduce more points of failure.

In this particular case, since we are introducing specificity based on classes and the element tag is an "article" (which is not super common), I feel like specifying is not bringing much and opening doors to errors with class changes.

But leaving it like that is good too. It's still simple and relying on just a few things ;)

processing/data_collection/gazette/spiders/rn_mossoro.py
giuliocc: Again, not too important, so I'll just wait for your closing thoughts.

Comparing to the last one, this introduces even more complexity. The time-related stuff seems to be enough.

processing/data_collection/gazette/spiders/rn_mossoro.py
giuliocc: I don't think the comment is necessary then. It goes without saying.

giuliocc:
Hi @vitorbaptista ! It happens hehehehe.

Right now I'll be giving preference to this PR since it was opened first and @phrfpeixoto is active on it. But I'll leave the other opened until merging this one.

Thanks for pointing that out :)

giuliocc:
One more thing: can we scrape the edition number since it is available? The field has been added recently and I feel like adding it right now would be a good opportunity hehhehe.

> @giuliocc Thanks for the review. I've commented and resolved the ones that I didn't agree, and left some that I'm open to change. Let me know your thoughts.

Thank you for the contribution :) Just made some comments and I'll be waiting for your closing thoughts.

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: Right. Changed.

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: Removed.

phrfpeixoto:
@giuliocc Solved your comments and added the `edition_number`

processing/data_collection/gazette/spiders/rn_mossoro.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def parse_gazette(self, response):
+        gazette_date = response.meta["date"]
+        file_url = response.css(".wp-block-file__button::attr(href)").get()
+        edition_text = response.css("h1.entry-title::text").get()
+        edition = re.search(r"JOM n\.º ([a-z0-9]+)$", edition_text, re.IGNORECASE)
+
+        yield Gazette(
+            date=gazette_date,
+            edition_number=edition and edition.group(1),
```
giuliocc, line 60:
```suggestion
        edition_regex = re.compile("JOM n\.º ([a-z0-9]+)$", re.IGNORECASE)
        edition = response.css("h1.entry-title::text").re_first(edition_regex)

        yield Gazette(
            date=gazette_date,
            edition_number=edition,
```

giuliocc (changes requested):
Almost done! Just one more suggestion, I promise :)

processing/data_collection/gazette/spiders/rn_mossoro.py
phrfpeixoto: Done.

jvanz (approved):

giuliocc:
Sorry for the delay :/ @jvanz already took care of it :)

Thank you for the PR and your patience, @phrfpeixoto . Very nice work!

Issue #260, opened 2020-10-02
Spider pr_curitiba is broken


giuliocc:
Apparently it works fine until the end, when it doesn't download the gazettes. When executed, it gives errors like this one:

```
2020-10-02 06:21:11 [scrapy.pipelines.files] WARNING: File (code: 301): Error downloading file from <GET http://legisladocexterno.curitiba.pr.gov.br/DiarioConsultaExterna_Download.aspx?id=4307> referred in <None>
2020-10-02 06:21:11 [scrapy.core.scraper] ERROR: Error processing {'date': datetime.date(2020, 9, 1),
 'file_urls': ['http://legisladocexterno.curitiba.pr.gov.br/DiarioConsultaExterna_Download.aspx?id=4307'],
 'files': [],
 'is_extra_edition': False,
 'power': 'executive_legislature',
 'scraped_at': datetime.datetime(2020, 10, 2, 6, 21, 11, 172065),
 'territory_id': '4106902'}
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/usr/local/lib/python3.7/site-packages/scrapy/utils/defer.py", line 157, in f
    return deferred_from_coro(coro_f(*coro_args, **coro_kwargs))
  File "/mnt/code/data_collection/gazette/pipelines.py", line 28, in process_item
    if self.is_doc(item["files"][0]["path"]):
IndexError: list index out of range
```

Issue #172, opened 2020-05-19
Create date filter


EvertonTomalok:
Hello guys, how we talked some days ago, we found a potential problem to integrate OKFN's projects to Jusbrasil stack of Brazilian data extraction.

We discussed some points about it, and we suggested some changes in the project:

- Create a way where is possible to pass a date to filter results.
- Based on these filters, make fewer requests than possible.
- [Minor] Return PDF's links and more information if possible, like publication date, etc.

An important point to remember is that we are available to help with coding theses changes, if necessary.

jvanz:
Hi Everton!

I'm sorry for the delay. I was stuck in a course that I was doing and could not reach my TODO list item to give you some feedback. Let's do it now:


>     * Create a way where is possible to pass a date to filter results.
>     * Based on these filters, make fewer requests than possible.

I would like to talk with you about this topic. You said you would like to avoid request, wich is good. But wanna to check if you want to avoid the requests to the web pages OR the request to download the files. As I said in the call, our current implementations visit all the pages. But it does not download all the files every time. We have [cache](https://github.com/okfn-brasil/diario-oficial/blob/master/processing/data_collection/gazette/settings.py#L11) in place witch prevent us to download a file already downloaded. Furthermore, we have a simple [date filter](https://github.com/okfn-brasil/diario-oficial/blob/master/processing/data_collection/gazette/pipelines.py#L43) already in use. This date filter is not working for the FECAM spiders. But I plan to add that soon to make it concise with the other spiders.

Both solutions are very simple. But it works so far. I would love to see a mechanism to avoid visit all the web pages. That will be awesome. We can talk about that more deeply.

>     * [Minor] Return PDF's links and more information if possible, like publication date, etc.

At the first glance, I believe this will be easy. Scrapy has way to export the scraped data. We can add something to export some file with the data instead of download the files in the disk.

giuliocc:
[BaseGazetteSpider](https://github.com/okfn-brasil/querido-diario/blob/d217c84df26a1547e18f6aad98d3f6f5ff7029c9/data_collection/gazette/spiders/base/__init__.py#L7) already has an `start_date` and `end_date` argument parsers now.

`start_date` in particular is being used in all spiders running in production and making as few requests as possible when a date is provided.

I'm closing this issue since I believe it's been fulfilled :)

PR #261, opened 2020-10-02
Alternative Date Filtering proposal

giuliocc:
This is motivated by #247 and #172 .

This middleware can filter requests based on the `meta` field `gazette_date` and incorporates the functionality of `GazetteDateFilteringPipeline`. So it can filter items too.

The `end_date` attribute was added to enable a more specific date range for the crawl. I guess this should benefit #172.

`pe_recife` and `pr_curitiba` were refactored just for demonstration purposes. If we were to merge, they could become other PRs.

If we were to merge, documentation should be added too, to indicate this practice for contributors.

By the way, `pr_curitiba` has an issue, #260 was opened for it.

### DIFF
--- a/processing/data_collection/gazette/middlewares.py
+++ b/processing/data_collection/gazette/middlewares.py
@@ -1,13 +1,72 @@
-# -*- coding: utf-8 -*-
-# Define here the models for your spider middleware
-#
-# See documentation in:
-# https://doc.scrapy.org/en/latest/topics/spider-middleware.html
-from scrapy import signals
+import datetime
+import logging
+
+from scrapy import Request, signals
+
+
+class GazetteDateFilteringSpiderMiddleware:
+    @classmethod
+    def from_crawler(cls, crawler):
+        s = cls()
+        s.logger = logging.getLogger(f"{__name__}.{cls.__name__}")
+        return s
+
+    def process_spider_output(self, response, result, spider):
+        for request_or_item in result:
+            if isinstance(request_or_item, Request):
+                if request_or_item.dont_filter is True:
+                    yield request_or_item
+                    continue
+
+                gazette_date = request_or_item.meta.get("gazette_date")
+                if gazette_date is None:
+                    yield request_or_item
+                    continue
+            else:
+                gazette_date = request_or_item["date"]
+
+            if not (
+                self.filtered_by_start_date(gazette_date, spider)
+                or self.filtered_by_end_date(gazette_date, spider)
+            ):
+                yield request_or_item
+            else:
+                self.logger.warning(
+                    f"Request or item outside of the desired date range was dropped: {request_or_item}"
+                )
+
+    def process_start_requests(self, start_requests, spider):
+        yield from self.process_spider_output(None, start_requests, spider)
+
+    def filtered_by_start_date(self, date, spider):
+        if not hasattr(spider, "start_date"):
+            return False
+
+        return spider.start_date > self.parse_date(date, spider.start_date)
+
+    def filtered_by_end_date(self, date, spider):
+        if not hasattr(spider, "end_date"):
+            return False
+
+        return spider.end_date < self.parse_date(date, spider.end_date)
+
+    def parse_date(self, date, default_date):
+        if isinstance(date, dict):
+            year = date.get("year")
+            month = date.get("month", default_date.month)
+            day = date.get("day", default_date.day)
+            parsed_date = datetime.date(year, month, day)
+        elif isinstance(date, datetime.date):
+            parsed_date = date
+        elif isinstance(date, datetime.datetime):
+            parsed_date = date.date()
+        else:
+            raise Exception(f"Date {date} cannot be parsed.")
+
+        return parsed_date
 
 
 class GazetteSpiderMiddleware:
-
     # Not all methods need to be defined. If a method is not defined,
     # scrapy acts as if the spider middleware does not modify the
     # passed objects.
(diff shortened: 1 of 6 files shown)


processing/data_collection/gazette/middlewares.py
```diff
@@ -1,13 +1,72 @@
...
+        return s
+
+    def process_spider_output(self, response, result, spider):
+        for request_or_item in result:
+            if isinstance(request_or_item, Request):
+                if request_or_item.dont_filter is True:
+                    yield request_or_item
+                    continue
+
+                gazette_date = request_or_item.meta.get("gazette_date")
```
rennerocha, line 21: This Middleware assumes that we know beforehand what dates the request will return. This is not always true. For example, in a paginated page, one request can return data from from different years/months/days. Also, if we can ensure that a request will return information from a specific date/period, as we have this information and `start_date` available in the spider, we could simply do the filtering in the spider, avoiding the use of a MIddleware that may or may not work in a more general basis.

processing/data_collection/gazette/pipelines.py
```diff
@@ -14,14 +13,6 @@
 from gazette.settings import FILES_STORE
 
 
-class GazetteDateFilteringPipeline:
```
rennerocha, line 17: I wouldn't remove this pipeline. This will always guarantee that we will return only the desired data, even if any other thing we include (like a Middleware or spider logic) fails.

processing/data_collection/gazette/spiders/base.py
```diff
@@ -9,14 +9,19 @@
...
+    def __init__(self, start_date=None, end_date=None, *args, **kwargs):
         super(BaseGazetteSpider, self).__init__(*args, **kwargs)
 
         if start_date is not None:
-            parsed_data = dateparser.parse(start_date)
+            parsed_data = dateparser.parse(start_date, settings={"DATE_ORDER": "DMY"})
             if parsed_data is not None:
                 self.start_date = parsed_data.date()
 
+        if end_date is not None:
```
rennerocha, line 20: Considering how we will run the spiders in production, I don't think we would need an `end_date`. Mainly we will get only the gazettes from the last day.

processing/data_collection/gazette/spiders/pe_recife.py
```diff
@@ -51,7 +51,7 @@ def start_requests(self):
         for date in dates:
             yield scrapy.Request(
                 url=self.EDITIONS_IN_DATE_URL.format(full_date=date.strftime("%Y%m%d")),
```
rennerocha, line 53: This spider is one the best spiders to filter by date as we have a one-to-one relation between a gazette and a request. So an easiest way to filter is just to create the `dates` list using `start_date` as `dtstart`. No need for Middlewares and the code will be much more clearer.

rennerocha (commented):
This Middleware certainly will filter the requests if they have available information of the response we are expecting (the dates that request will provide). However, for websites that have pagination, this is not always an easy thing to identify. I tend to prefer to keep the filtering inside each spider and avoid middlewares that may have invisible side effects that may be difficult to find considering that we don't have the spiders running now. I changed two spiders (`rn_natal` and `sc_florianopolis`) to filter by date in the spider and the changes weren't that big, so I think it will even be easier for new developments to keep that way and we update the existing spiders to follow them.

processing/data_collection/gazette/middlewares.py
giuliocc: > This Middleware assumes that we know beforehand what dates the request will return. This is not always true. For example, in a paginated page, one request can return data from from different years/months/days.

Correct me if I made a mistake as I'm not sure I understood this particular concern. But this implementation allows not filtering requests for not having the date information and when `dont_filter` in present in the meta. So, we could agree that filtering with the middleware and filtering inside the spider would provide the same functionality?

Also, it allows partial dates like "only having the year" or "only having year and month".

For pagination, using `sp_jundiai` as an example, information about the gazettes dates could be extracted and used in the requests for the gazette info page and also for the pagination if we used the date of the last displayed gazette in the pagination request (since the system shows gazettes in reverse sorted manner, this shouldn't be a problem). Also, we could not give any info about the date in the pagination requests, if that logic is not correct. I think this implementation would be pretty similar to what we would be able to do inside the spider, right? For others systems that don't show this information, we can't do much about that either way :/

> Also, if we can ensure that a request will return information from a specific date/period, as we have this information and start_date available in the spider, we could simply do the filtering in the spider, avoiding the use of a MIddleware that may or may not work in a more general basis.

If the functionality of doing it inside the spiders is the same, why wouldn't it be better to make a middleware responsible for this? Just like the `GazetteDateFilteringPipeline` is responsible for filtering the items instead of writing it in every spider, don't you think we should centralize it?

If we were to add another argument like `end_date` or something we haven't thought about right now, all spiders would have to be refactored again if that code belonged in every spider, right?

processing/data_collection/gazette/pipelines.py
giuliocc: That was removed because it was basically doing the same thing as the spider middleware for the items

processing/data_collection/gazette/spiders/base.py
giuliocc: Ok. But that provides ways to rerun dates that failed with `RetryError` or something like that. Don't you think that would be desirable?

processing/data_collection/gazette/spiders/pe_recife.py
giuliocc: I kind of agree here. But consider my point about adding some other argument like `end_date` or something else we haven't thought about and having to refactor every spider again.

rennerocha:
`start_date` and `end_date` attributes were added and we should use them to filter the gazettes returned. A pipeline exists that drops the items that are not withing the desired range, however, the spiders are also responsible for doing the filtering to avoid unnecessary requests.

PR #262, opened 2020-10-02
Add all cities

alexandrevicenzi:


alexandrevicenzi:
The idea here is to add all cities so people can know what is implemented or not, this may be a good start if someone wants to help. The previous list was a bit outdated and had some issues.

Let me know if it would be nice to add more information to the table.

ejulio:
Awesome!
I'd just push the script you probably used to create it so people won't need to update it manually.

alexandrevicenzi:
@ejulio where it would fit (path)? the script is also a mess, so I need to work properly on it.

ejulio (commented):
For me, this PR is good to go.
Not sure about the path for the script and it can be added in a new PR if that makes sense.
Though, I think it wouldn't be a problem to add it here..

@rennerocha thoughts on it?

jvanz (changes requested):
I think the script can live in some directory like `devel` or `scripts`. Furthermore, add a target in the Makefile. Thus, the user does not need to know how to call the script . Something like `make update_cities`.

Please, rebase your branch. There is some conflict with the latest changes on main. Thanks!

alexandrevicenzi:
I'll rebase and add the script + command on makefile.

giuliocc:
I think this idea is great :) Nice work!

anapaulagomes:
You have also the option of adding the script as a scheduled script on Github Actions, so it would run and update the list daily, for instance.

jvanz:
> You have also the option of adding the script as a scheduled script on Github Actions, so it would run and update the list daily, for instance.

That's interesting idea! We could run this when we got a pull request merged.

rennerocha:
Keeping cities manually updated is not scaling. Probably we will use a database to store this information and deprecate CITIES.md file in future.

Issue #248, opened 2020-09-30
Add spider monitoring


rennerocha:
During the discussions of #247 a suggestion was made to verify the ratio `number_of_requests / items_scraped`. A higher ratio may be a sign that the spider is doing more requests than necessary and it can be a performance issue.

[Spidermon](https://spidermon.readthedocs.io/en/latest/) is an extension designed for spider monitoring that allows us to create custom monitors that will be executed every time a spider is executed.

It will be interesting to include this monitoring tool to the project and later, when the project is in production, add more checks that would ensure that everything will be working fine.

Suggested steps for this issue:
- [ ] Add Spidermon as project dependency and enable it in the project
- [ ] Create a new Monitor Suite
- [ ] Create a new monitor that will check the ratio `number_of_requests / items_scraped`. This monitor will fail if the ratio is higher than a configurable threshold (so we can define different threshold for different spiders if needed)

rennerocha:
@andrewmiko 

If you have any problem configuring Spidermon just let me know. Following Spidermon [tutorial](https://spidermon.readthedocs.io/en/latest/getting-started.html) should be enough for you to understand how to enable it and create one basic monitor.

For this issue, it is not necessary yet to define actions to be taken when the monitor fail. This will be decided later when we have spiders running in production.

andrewmiko:
OK. Sounds very good. Please assign this issue to me. I'll try to work on it later this week. 👍

andersonberg:
Is this issue being handled yet? If not, can I start it?

andrewmiko:
@andersonberg I think that you can take this one. Currently, I'm busy in other projects. Cheers.

andersonberg:
Thank you, @andrewmiko

rennerocha:
Closed by #304

PR #263, opened 2020-10-02
TERRITORY_ID should be a string

alexandrevicenzi:
All other files use it as string and to avoid errors it's best to use
the same pattern everywhere.


rennerocha:
Thanks for noticing that. It is a small change but it is better to be consistent.

rennerocha:
#248 will introduce an extension that will allow us to monitor if we are using the wrong data type on this field.

ejulio (approved):

PR #268, opened 2020-10-03
Validate format of start_date argument

rennerocha:
Having a fixed pattern for this argument seems better when we start
scheduling hundreds of spiders daily. It also removes the dependency on
dateparser (that can have performance issues depending the format you
are trying to parse).


ejulio (approved):
:ok_hand:

PR #269, opened 2020-10-04
Finish Governador Valadares/MG spider

giovanisleite:
This is part of the efforts needed for #221
Closes #19 

- Refactor the parse method
- Changed the request headers to meet only what is needed
- Updated the city URL (Add https and the endpoint changed too)
- Get the URL dynamically (At the start of this PR the URL was another)
- Filter by start date

I was able to download all files:

image
```bash
$ ls data/full/*.txt | wc -l
1624
```


processing/data_collection/gazette/spiders/mg_governador_valadares.py
```diff
@@ -0,0 +1,94 @@
...
+
+        self.current_page += 1
+        yield self.make_request(self.start_urls[0], self.current_page)
+
+    def make_body(self, page):
+        return json.dumps(
+            {
+                "Page": page,
+                "cdCaderno": 1,
+                "Size": str(self.ITEMS_PER_PAGE),
```
jvanz, line 82: You use this variable just in one place. Why not use a string value by default:? Thus, you do not need to convert it into a string. What do you think?

processing/data_collection/gazette/spiders/mg_governador_valadares.py
```diff
@@ -0,0 +1,94 @@
...
+        url = self.start_urls[0]
+        yield self.make_request(url, self.current_page)
+
+    def make_request(self, url, page):
+        return scrapy.Request(
+            url,
+            method="POST",
+            headers={"X-AjaxPro-Method": "GetDiario",},
+            body=self.make_body(page),
+            dont_filter=True,
```
jvanz, line 33: Is there a specific reason of why the request cannot be filtered?

jvanz (changes requested):

processing/data_collection/gazette/spiders/mg_governador_valadares.py
giovanisleite: I think the author used it because it is a lot of requests to the same URL, but I don't know if it is needed. I will look into it

processing/data_collection/gazette/spiders/mg_governador_valadares.py
giovanisleite: Done! Thanks!

processing/data_collection/gazette/spiders/mg_governador_valadares.py
```diff
@@ -0,0 +1,113 @@
...
+        )[0]
+        content = content.replace("new Date", "")
+
+        rows = None
+        definition = None
+
+        try:
+            definition, rows = self.extract_definitions_and_rows(content)
+        except Exception as e:
+            return
```
jvanz, line 49: I would move this code to another function with a meaningful name.

processing/data_collection/gazette/spiders/mg_governador_valadares.py
```diff
@@ -0,0 +1,113 @@
...
+    def parse_items(self, response):
+        body = response.body
+        has_no_results = body == "null;/*".encode()
+        if has_no_results:
+            return
+
+        content = re.findall(
+            "new Ajax\.Web\.DataTable\((?P<conteudo>.*)\);", body.decode("utf-8")
+        )[0]
+        content = content.replace("new Date", "")
```
jvanz, line 41: I would move this code to another function with a meaningful name.

processing/data_collection/gazette/spiders/mg_governador_valadares.py
```diff
@@ -0,0 +1,113 @@
...
+        scripts = response.xpath("//script//@src").extract()
+        endpoint = next(path for path in scripts if "diel_diel_lis" in path)
+        self.path = endpoint
+        yield self.make_request(self.current_page)
+
+    def parse_items(self, response):
+        body = response.body
+        has_no_results = body == "null;/*".encode()
+        if has_no_results:
+            return
```
jvanz, line 36: I would move this code to another function with a meaningful name.

jvanz (commented):
I do not see nothing bad. But I would just extract some lines into functions with explanatory names.

processing/data_collection/gazette/spiders/mg_governador_valadares.py
giovanisleite: This is part of `parse_definitions_and_rows` now.

processing/data_collection/gazette/spiders/mg_governador_valadares.py
giovanisleite: This is part of `extract_python_expression` now

processing/data_collection/gazette/spiders/mg_governador_valadares.py
giovanisleite: This is part of `is_body_empty` now

processing/data_collection/gazette/spiders/mg_governador_valadares.py
```diff
@@ -0,0 +1,120 @@
...
+    ITEMS_PER_PAGE = "100"
+    BASE_URL = "https://www.valadares.mg.gov.br/"
+
+    path = ""
+    current_page = 0
+
+    def parse(self, response):
+        scripts = response.xpath("//script//@src").extract()
+        endpoint = next(path for path in scripts if "diel_diel_lis" in path)
+        self.path = endpoint
```
jvanz, line 29: This is not wrong. But I would say to move this to another function with a good name. Maybe, something like `get_script_path_to_request_items`. I'm not sure this a good name, but I think something like this will give us a even better code!

jvanz:
Sorry for the delay... =/

I've just run the spider. It works fine. Awesome work @giovanisleite ! Thanks for your help one this!

giovanisleite:
> I do not see nothing bad. But I would just extract some lines into functions with explanatory names.

Done!

Thank you for your help, @jvanz

jvanz (approved):

PR #252, opened 2020-10-01
Automatically sort imports using isort module (#241)

jaswdr:
Sort all imports by using `isort` library with options to not conflict with existing `black` library.

Signed-off-by: Jonathan Schweder <jonathanschweder@gmail.com>

Issue #148, opened 2019-12-12
spider automation test


jvanz:
I believe we should find a way to run some kind of automatic testing in the spiders. My first idea is use the github actions to run the spiders for a while and check if it downloads some files. If it does, great, the test passes. Otherwise, it fails. 

Furthermore, if the test failed, we can use the github action to open an issue automatically.

vitorbaptista:
For the parsers, I like the way @pauloabelha and I wrote the parser on https://github.com/okfn-brasil/diario-oficial/pull/106 following a TDD style.

For the spiders, I think we'd be better off monitoring than testing.

Let's take a step back. The problem we're facing is how to ensure that the scrapers are healthy and downloading the data as normal. For example, how could we notice if a public body changed its gazette format?

We could solve it using tests. There are a few tools that help saving and replaying an HTML response (personally I like [VCR.py](https://github.com/kevin1024/vcrpy)). We could write a test that downloads a gazette's webpage and parses it, checking the results. In this case, we should also write a test that downloads the *latest* gazette and tries to parse it, as I imagine a common change will be a new gazette template.

This would work, and a test suite can be very useful while developing the scraper (if it's easy enough to set up).

Another possibility is simply to add monitoring to our scrapers. If I expect 100 scrapers to download 100 new gazettes per day, if one day only 95 gazettes are downloaded, I know something went wrong. This IMHO would be an easier way to increase the pipeline's reliability.

Just to be clear, ideally we would have both: testing and monitoring. Testing focusing more on writing the scrapers, TDD-style (as in #106), while monitoring focus on checking if the pipeline is healthy and nothing broke because of i.e. a website structure change.

For monitoring, I'd suggest taking a look at https://www.datadoghq.com/. For testing, VCR-py is great, and if we follow the [Functional Core, Imperative Shell](https://www.destroyallsoftware.com/screencasts/catalog/functional-core-imperative-shell) pattern, we could just save HTMLs and pass them on to be parsed.

rennerocha:
Testing spiders is not an easy task as you have absolutely no control over the content you receive from the websites you are scraping so we can't think of tests the same way as other systems.

What can be tested? Processors and pipelines are good candidates for regular testing. Parsing methods don't.

Caching the responses (using VCR.py or any other kind of library that stores it) may be useful only for development, and I don't see any real gain to keep it in the project, as they are very fragile tests. Any change in the website layout, the tests will be outdated and useless.

[...]

PR #276, opened 2020-10-06
Inserted Guidelines to maintainers

sergiomario:
I could open an issue to discuss this, but I have adopted a philosophy of starting a discussion with a proposal.

I created this chapter to propose an organization for the project maintainers. The idea is that it is not legislation with multiple responsibilities, but a guide that facilitates communication and reduces rework.

I do this when we invite @giuliocc to compose our team of maintainers. :1st_place_medal:  Thus, we are growing and creating a working model to prevent distress to people as generous as Querido Diario maintainers.

"There is a criteria I call change latency, which is the round-trip time from identifying a problem to testing a solution. The faster the better. If maintainers cannot respond to pull requests as rapidly as people expect, they're not doing their job (or they need more hands)." (Peter Hintjens)
https://hintjens.gitbooks.io/social-architecture/content/chapter4.html

I would like opinions: @jvanz @rennerocha

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,8 +60,6 @@ This section is a space for project maintainers, since as code review and reposi
 ### Responsibilities of a Querido Diario maintainer:
 
 - Respect the [code of conduct](https://github.com/okfn-brasil/querido-diario/blob/main/CODE_OF_CONDUCT.md) and ensure that anyone who may have suffered harassment has a help channel.
-- Manage the labels to avoid rework and conflict communication. ([CPython project](https://github.com/python/cpython/pulls) as a reference)
-    - *awaiting review*
-    - *awaiting changes*
-    - *awaiting merge*
 - Always justify a suggestion according to the practices already adopted in the project, legibility and simplicity. It is essential for a civil project to have a structure as easy as possible for beginners.
+- The maintainers should **run all the spiders before merging it**.
+- If a Pull Request have many commits and their messages are not clear, use the *Squash and merge* option. It is not necessary to do squash when the commits discipline is good, as an example: [Pull Request](https://github.com/okfn-brasil/querido-diario/pull/252/commits).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,3 +52,16 @@ $ black .
 ```
 
 It is important to note that the CI will fail if you commit Python code that is not in accordance with Black code style.
+
+## Guidelines to maintainers
+
+This section is a space for project maintainers, since as code review and repository organization are part of the contribution process, all code review guidelines are of public interest to anyone in the community.
+
+### Responsibilities of a Querido Diario maintainer:
+
+- Respect the [code of conduct](https://github.com/okfn-brasil/querido-diario/blob/main/CODE_OF_CONDUCT.md) and ensure that anyone who may have suffered harassment has a help channel.
+- Manage the labels to avoid rework and conflict communication. ([CPython project](https://github.com/python/cpython/pulls) as a reference)
+    - *awaiting review*
+    - *awaiting changes*
+    - *awaiting merge*
+- Always justify a suggestion according to the practices already adopted in the project, legibility and simplicity. It is essential for a civil project to have a structure as easy as possible for beginners.


jvanz:
I would like to raise two topics for discussion:

1. Squash commits: My personal preference is to have a clean git history. Thus, when we have a git branch with a lot of commits fixing review issues, we should remember to squash the commit. Leaving the meaningful ones only.
 
2. Should we require more then one reviewer? I do not know if this is realistic for the current state of the project. But I think we should give opportunity to the community to review other members PRs. This can be a soft rule. It does not need to be a hard rule to get the PR merged.

rennerocha:
I don't think that Squash all PRs is the best policy. It is a good idea when we are adding a brand new spider but sometimes, if the contributor has a good discipline, the commits can be separately well defined. Probably it may be a soft rule, but when we are merging we can choose just to rebase if we consider that the commit history is good as it is.

As an example I can mention this [PR](https://github.com/okfn-brasil/querido-diario/pull/252/commits). One commit is adding the formatting configurations and a second commit is applying the formatting into the codebase. I'd prefer to have a separated commit for this because if I want to understand when automatic formatting was added, it will be a mess to look into all changes in the codebase until I find what I am looking for.

rennerocha:
>     2. Should we require more then one reviewer? I do not know if this is realistic for the current state of the project. But I think we should give opportunity to the community to review other members PRs. This can be a soft rule. It does not need to be a hard rule to get the PR merged.

I don't think we need this for now.

rennerocha:
>     3. The maintainers should run all the spiders before merging it.

Agreed!!! I reviewed several spiders that the contributor certainly didn't run the spider entirely once.

rennerocha:
@sergiomario 
I didn't understand the labeling rules very well. In my opinion it will be good to have a well defined set of labels and documented rules when to use them.

jvanz:
> > ```
> > 1. Squash commits: My personal preference is to have a clean git history. Thus, when we have a git branch with a lot of commits fixing review issues, we should remember to squash the commit. Leaving the meaningful ones only.
> > ```
> 
> I don't think that Squash all PRs is the best policy. It is a good idea when we are adding a brand new spider but sometimes, if the contributor has a good discipline, the commits can be separately well defined. Probably it may be a soft rule, but when we are merging we can choose just to rebase if we consider that the commit history is good as it is.
> 
> As an example I can mention this [PR](https://github.com/okfn-brasil/querido-diario/pull/252/commits). One commit is adding the formatting configurations and a second commit is applying the formatting into the codebase. I'd prefer to have a separated commit for this because if I want to understand when automatic formatting was added, it will be a mess to look into all changes in the codebase until I find what I am looking for.

I agree with you. Sorry, I would not  sound like "squash everything". As I said `Leaving the meaningful ones only.`  I just wanna avoid is like "Fix typo1", "Fix typo2", "Fix typo3" or a generic "Review fixes". Any way, I think we are in the same page.

rougeth:
Hey folks, I was reading your comments and found this topic super relevante/interesting. I have one question about @jvanz's comment:

> The maintainers should run all the spiders before merging it.

It should be possible to run the spider in a Github Action (or other CIs), right? Github API allows to see which files were changed in a Pull Request, so there's a way to know which spiders were changed. What do you think about it?

rennerocha:
> It should be possible to run at least part of the spider that validates the code change in a Github Action, right?

Is it? I have no idea :smile: Never worked with Github Actions so I don't know if this is possible or not. But it would be a great improvement if we can setup a basic environment with Scrapy that runs the spider that changed with files download disabled and check some basics stats (if items were scraped, if it has no errors, etc). Maybe Travis can do something like that?

Anyway, asking for the maintainers to run the spiders locally as well is still a good advice ...

jvanz:
Yeah, I think is possible to run the spider in Github Actions. That's my idea when I created the issue #148. But I think that I was not clear enough. We can try again. Open a issue for that with more info about what we want and how to consider a spider okay or not. ;-)

sergiomario:
Hi firends. :)

I updated the guidelines.
- Removed labels control - I think that as funcionalidades de fluxo de revisão de código que existem no Github resolvem os problemas que eu havia pensado em resolver com as labels de controle.
- "The maintainers should run all the spiders before merging it." @jvanz :+1: 
- Squash and merge with the caveats described by @rennerocha

jvanz (approved):

rennerocha (approved):

PR #277, opened 2020-10-06
Add edition number field

rennerocha:
Practically all cities provides the edition number of gazettes and we weren't collecting it.
As usually this is something easy to collect and it seems useful to be able to find a gazette with another property, we should start collecting it. 
I updated just one spider for testing, but we will need to update all spiders to include this field.


giuliocc (approved):
Florianópolis is working as expected. And I agree that using this attribute as a string is the way to go.

LGTM.

giuliocc:
The CI check failed but that was unrelated to this PR.

PR #280, opened 2020-10-07
Guia de desenvolvimento de raspadores | Spider development guidelines

rennerocha:
This document is intended to be a set of good practices and useful information that a developer should follow when developing a new spider or updating an existing spider. The idea is to be a more technical document that contains patterns to follow, things to be aware of and good practices.

Comments? Suggestions of more information that may be included here?

### DIFF
--- a/SPIDER-GUIDELINES.md
+++ b/SPIDER-GUIDELINES.md
@@ -1,0 +1,70 @@
+Spider writing guidelines
+=========================
+
+Before start working on a spider, read carefully the following guidelines that will make the project more consistent and your PR more probable to be accepted.
+
+Reading code of existing spiders will also help you to understand how to develop.
+
+Spider class
+------------
+
+- All spiders must inherit from `BaseGazetteSpider`
+- Class name must follow the pattern `<State><CityName>Spider` (for example `SpSaoPauloSpider` or `MtFelizNatalSpider`)
+- Spider `name` must follow the pattern `<state>_<city_name>` (for example `sp_sao_paulo` or `mt_feliz_natal`)
+- Add `TERRITORY_ID` attribute with the code found in `territories.csv` file
+- Add `start_date` attribute containing the date of the first gazette available
+
+Example initial spider:
+```
+from datetime import date
+from gazette.spiders.base import BaseGazetteSpider
+
+class SpJundiaiSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3525904"
+    name = "sp_jundiai"
+    allowed_domains = ["jundiai.sp.gov.br"]
+    start_urls = ["https://imprensaoficial.jundiai.sp.gov.br/"]
+    start_date = date(2000, 1, 1)
+
+    def parse(self, response):
+        # Here we have all parsing logic
+```
+
+Field definitions
+-----------------
+
+These are the information that we need to extract for each gazette:
+
+- **date (datetime.date)**: publication date
+- **edition_number (string)**: edition number of gazette
+- **is_extra_edition (boolean)**: some gazettes are extra editions. You can identify them when they have words like *Extra*, *Extraordinário*, *Suplemento*, etc. If unable to identify, return as `False`
+- **territory_id (string)**: value of `TERRITORY_ID` spider attribute
+- **power (string)**: if the gazette refers to Executive power, Legislative or both (`executive`, `legislative`, `executive_legislative`)
+- **scraped_at (datetime.datetime)**: fixed to `datetime.datetime.utcnow()`
+- **file_urls (URL list)**: list of URLs to download the gazettes (usually we will have just one value here)
+
+Date filtering
+--------------
+
+We want all spiders to be able to retrieve **ALL** available gazettes. But we also want to be able to restrict the amount of data we retrieve. To achieve that we have `start_date` attribute that will help filtering.
+
+By default this attribute must have the date of the first gazette available in the site (hard-coded). But we can change that between executions passing the desired date as argument. The following command should return only gazettes after 01/Sep/2020:
+
+`scrapy crawl my_spider -a start_date=2020-09-01`
+
+When developing a spider, try to make **as few requests as possible**. Some pages allows you to filter the results passing date information, so use `start_date` information in your favour.
+
+Before submitting a PR
+----------------------
+
+**RUN THE SPIDER IN YOUR DEVELOPMENT ENVIRONMENT AND WAIT UNTIL IT FINISHES ITS EXECUTION!**
+
+Before submitting a new PR for your spider, make sure that it is running properly:
+
+- Check if the downloaded files are valid;
+- Check if you are getting all expected gazettes;
+- Check if when you specify a `start_date` you are not collecting more gazettes than expected;
+- Check if you run the spider twice, the results are the same;
+- Check for ERRORS in spider stats
+- ...
+


sergiomario:
@rennerocha this is an excellent idea! Will help us a lot in tech events with beginners.

I have only one question: I'm trying to create a [maintainers guidelines chapter in the CONTRIBUTING.md file](#276).
Do you think that all guidelines should come to this new file or is it an idea to just focus on spiders? If it is just for spider creation, could the name be `SPIDERS-GUIDELINES.md`?

giuliocc:
I liked it very much!

I'm thinking that a link from `CONTRIBUTING.md` to this file is necessary. And for extra content maybe some tips? Like disabling download of files to accelerate some parts of the development.

GUIDELINES.md
```diff
@@ -0,0 +1,70 @@
...
+=========================
+
+Before start working on a spider, read carefully the following guidelines that will make the project more consistent and your PR more probable to be accepted.
+
+Reading code of existing spiders will also help you to understand how to develop.
+
+Spider class
+------------
+
+- All spiders must inherit from `BaseGazetteSpider`
```
alexandrevicenzi, line 11: This is not entirely true, `FecamGazetteSpider` is a base class that can be used and someone could also create a new base spider, it may be relevant to note that.

Of course, all spiders and base spiders have to inherit from `BaseGazetteSpider` at some point.

GUIDELINES.md
```diff
@@ -0,0 +1,70 @@
...
+    start_date = date(2000, 1, 1)
+
+    def parse(self, response):
+        # Here we have all parsing logic
+```
+
+Field definitions
+-----------------
+
+These are the information that we need to extract for each gazette:
```
alexandrevicenzi, line 36: Perhaps add an example of `Gazette` definition.

rennerocha:
> @rennerocha this is an excellent idea! Will help us a lot in tech events with beginners.
> 
> I have only one question: I'm trying to create a [maintainers guidelines chapter in the CONTRIBUTING.md file](#276).
> Do you think that all guidelines should come to this new file or is it an idea to just focus on spiders? If it is just for spider creation, could the name be `SPIDERS-GUIDELINES.md`?

I think that include everything in CONTRIBUTING will make that file too big. This guideline is more intended to be a reference for spider development only, so I believe a separated file suits better. I agree that the filename wasn't good. Your suggestion is much better! It is renamed.

GUIDELINES.md
```diff
@@ -0,0 +1,70 @@
...
+By default this attribute must have the date of the first gazette available in the site (hard-coded). But we can change that between executions passing the desired date as argument. The following command should return only gazettes after 01/Sep/2020:
+
+`scrapy crawl my_spider -a start_date=2020-09-01`
+
+When developing a spider, try to make **as few requests as possible**. Some pages allows you to filter the results passing date information, so use `start_date` information in your favour.
+
+Before submitting a PR
+----------------------
+
+**RUN THE SPIDER IN YOUR DEVELOPMENT ENVIRONMENT AND WAIT UNTIL IT FINISHES ITS EXECUTION!**
```
jvanz, line 60: I think it is worth to mentioned that the spider *may* send a lot of request to the web sites. Usually, the sites do not like that. I think we should, at least, put a link to show the user how to control the requests rates. Thus, they will not send gazillions of request at the same time. I'm mentioning that because I've already receive a call from my ISP telling that I was doing a DOS in a web site while running my spiders.

anapaulagomes: A way of preventing this is adding cache to the project or enabling autothrottle. What are your thoughts on this?

GUIDELINES.md
```diff
@@ -0,0 +1,70 @@
+Spider writing guidelines
+=========================
+
+Before start working on a spider, read carefully the following guidelines that will make the project more consistent and your PR more probable to be accepted.
+
+Reading code of existing spiders will also help you to understand how to develop.
```
anapaulagomes, line 6: What about adding a link to this?

```suggestion
Reading code of [existing spiders](https://github.com/okfn-brasil/querido-diario/tree/main/processing/data_collection/gazette/spiders) will also help you to understand how to develop.
```

It may change but sometimes it isn't clear where to look when one is starting in a new project.

GUIDELINES.md
```diff
@@ -0,0 +1,70 @@
...
+
+Spider class
+------------
+
+- All spiders must inherit from `BaseGazetteSpider`
+- Class name must follow the pattern `<State><CityName>Spider` (for example `SpSaoPauloSpider` or `MtFelizNatalSpider`)
+- Spider `name` must follow the pattern `<state>_<city_name>` (for example `sp_sao_paulo` or `mt_feliz_natal`)
+- Add `TERRITORY_ID` attribute with the code found in `territories.csv` file
```
anapaulagomes, line 14:
```suggestion
- Add `TERRITORY_ID` attribute with the code found in `data/territories.csv` file
```

Cool! I love the built-in search provided by Github for csv files. I think we mentioned a Wikipedia page somewhere... maybe we should update it to this file as well.

GUIDELINES.md
anapaulagomes, line 11:
```suggestion
- All spiders must inherit from `BaseGazetteSpider` or from a class that inherits from it
```

GUIDELINES.md
```diff
@@ -0,0 +1,70 @@
...
+
+Field definitions
+-----------------
+
+These are the information that we need to extract for each gazette:
+
+- **date (datetime.date)**: publication date
+- **edition_number (string)**: edition number of gazette
+- **is_extra_edition (boolean)**: some gazettes are extra editions. You can identify them when they have words like *Extra*, *Extraordinário*, *Suplemento*, etc. If unable to identify, return as `False`
+- **territory_id (string)**: value of `TERRITORY_ID` spider attribute
```
anapaulagomes, line 41: Would be nice to mention that this field is mandatory. Also, what do you think of mentioning that if no start date is informed is understood that all gazettes will be collected?

anapaulagomes (changes requested):
Having a spider guideline is indeed a really nice touch for this project. 🥇 Nice work!

I think it is worth mentioning useful commands from Scrapy such us `shell` and `-o` for exporting the scraped data. They are essential when developing new spiders.

GUIDELINES.md
jvanz: Yeah, I agree. I would say a link to the scrapy autothrottle documentation is sufficient. But if we decide to add a default autothrottle is great!

GUIDELINES.md
rennerocha: This is true, but I believe this is a more advanced use of base classes, so only more advanced contributors will worry about that.

GUIDELINES.md
rennerocha: Usually developing a spider will not schedule too many requests. If we notice that this is a problem (if contributors starts complaining about this problem or we have this problem in production running spiders), we can include that information.

GUIDELINES.md
rennerocha: In my opinion, adding links to files in the repository is prone to errors. If we move the file, it is very likely that we will forget to update the docs.

GUIDELINES.md
rennerocha: All fields are mandatory (except edition_number). I will add a note about that.
About start_date, you are right! It will be good to mention that.

SPIDER-GUIDELINES.md
```diff
@@ -0,0 +1,70 @@
...
+-----------------
+
+These are the information that we need to extract for each gazette:
+
+- **date (datetime.date)**: publication date
+- **edition_number (string)**: edition number of gazette
+- **is_extra_edition (boolean)**: some gazettes are extra editions. You can identify them when they have words like *Extra*, *Extraordinário*, *Suplemento*, etc. If unable to identify, return as `False`
+- **territory_id (string)**: value of `TERRITORY_ID` spider attribute
+- **power (string)**: if the gazette refers to Executive power, Legislative or both (`executive`, `legislative`, `executive_legislative`)
+- **scraped_at (datetime.datetime)**: fixed to `datetime.datetime.utcnow()`
```
jvanz, line 43: I'm wondering, could not add this in the same code where we set the territory_id automatically?

SPIDER-GUIDELINES.md
```diff
@@ -0,0 +1,70 @@
...
+
+Field definitions
+-----------------
+
+These are the information that we need to extract for each gazette:
+
+- **date (datetime.date)**: publication date
+- **edition_number (string)**: edition number of gazette
+- **is_extra_edition (boolean)**: some gazettes are extra editions. You can identify them when they have words like *Extra*, *Extraordinário*, *Suplemento*, etc. If unable to identify, return as `False`
+- **territory_id (string)**: value of `TERRITORY_ID` spider attribute
```
jvanz, line 41: Don't we have a piece of code which set this automatically?

jvanz (commented):
I have some questions...

SPIDER-GUIDELINES.md
giuliocc: It already does :)

rennerocha:
Moved to https://github.com/okfn-brasil/querido-diario-comunidade/pull/12

Issue #281, opened 2020-10-08
Petrópolis spider


Jefersonalves:
Add spider for Petrópolis/RJ

PR #282, opened 2020-10-08
Add spider for Petrópolis/RJ

Jefersonalves:
This PR solves #281


processing/data_collection/gazette/spiders/rj_petropolis.py
```diff
@@ -0,0 +1,61 @@
...
+            entry_url = entry.css("a::attr(href)").get()
+            if entry_url not in self.set_to_discard:
+                url = urljoin(self.BASE_URL, entry_url)
+                if re.search(r"\d+-\d+", url):
+                    yield scrapy.Request(url=url, callback=self.parse)
+                else:
+                    yield scrapy.Request(url=url, callback=self.parse_items_page)
+
+    def parse_items_page(self, response):
+        for entry in response.css("#col1 > div > table"):
```
alexandrevicenzi, line 38: You could get URLs using the selector `jd_download_url`.

processing/data_collection/gazette/spiders/rj_petropolis.py
```diff
@@ -0,0 +1,61 @@
...
+class RjPetropolis(BaseGazetteSpider):
+    TERRITORY_ID = "3303906"
+    BASE_URL = "http://www.petropolis.rj.gov.br"
+
+    name = "rj_petropolis"
+    allowed_domains = ["petropolis.rj.gov.br"]
+    start_urls = [
+        "http://www.petropolis.rj.gov.br/pmp/index.php/servicos-na-web/informacoes/diario-oficial/viewcategory/3-diario-oficial.html"
+    ]
+    set_to_discard = {
```
alexandrevicenzi, line 22: this seems like a hack because the implementation is not perfect or because the selectors are too broad.

alexandrevicenzi (changes requested):
I don't find this implementation clean or easy to understand, it seems that the code is trying to do too much, in too little code.

Petropolis uses a directory-based style to show the data that we want. I would expect the code to follow the same as:

- Follow Year directory
- Follow Month directory
- Parse Month page and yield all PDF links


The current code just looks for all links found in a broad way and use regex to figure out what function to call, this seems odd to me.

Jefersonalves:
Hi @alexandrevicenzi. Thanks for the review. I made some changes to make the spider simpler, as you suggested. I think this PR is ready for a new review

jvanz (approved):

giuliocc (commented):
The spider looks great! May I add just one little suggestion? Renaming the variables from `entry` to something like `year`, `month`, `gazette`, `document`, etc. Just to make it clearer when reading the code?

Jefersonalves:
> The spider looks great! May I add just one little suggestion? Renaming the variables from `entry` to something like `year`, `month`, `gazette`, `document`, etc. Just to make it clearer when reading the code?

@giuliocc Done!

jvanz (changes requested):
The code looks good. But I would like to take the opportunity to start make the spider more efficient. We are in the process to adjust the spiders to avoid unnecessary requests.  To me, you should change the spider to consider the `start_date` when walking in the pages. This variable can be change by the users to define witch gazeetes they want. So, my suggestion is to check if you found a year which is not request, in other words, is before the `start_date`, ignore it. The same for the months and days. Therefore, your spider will be more efficient and faster.

Jefersonalves:
@jvanz the spider is now more efficient and considers the start date to reduce the number of requests. Travis is failing because of the formatting of the settings.py file in the main branch

processing/data_collection/gazette/spiders/rj_petropolis.py
```diff
@@ -0,0 +1,96 @@
...
+        "abril": 4,
+        "maio": 5,
+        "junho": 6,
+        "julho": 7,
+        "agosto": 8,
+        "setembro": 9,
+        "outubro": 10,
+        "novembro": 11,
+        "dezembro": 12,
+    }
```
jvanz, line 50: I think you should use the standard library for this:

```
➜  querido-diario git:(main) ✗ python3 test.py
janeiro
fevereiro
março
abril
maio
junho
julho
agosto
setembro
outubro
novembro
dezembro
➜  querido-diario git:(main) ✗ cat test.py 
import calendar

with calendar.different_locale("pt_BR.utf8") as different_locale:
    for number in range(1,13):
        print(calendar.month_name[number])
```

;-)

jvanz (changes requested):

rennerocha:
Closed as stale.

PR #284, opened 2020-10-09
Database storage

rennerocha:
Considering that in production we will have to store scraped items information in a database, we need a pipeline that do this.

[Previously](https://github.com/okfn-brasil/querido-diario/pull/220) PostgreSQL dependency was removed to make it easier for development (as we won't need to setup a PostgreSQL container beforehand). However we still need to be able to store things in a database.

This PR will reintroduce the removed pipeline and, as we are using SQLAlchemy to define database schema, we can set a local SQLite as the location to store data extracted when developing. The main benefit with this approach is that we keep running the database pipeline at development time but we don't require any extra configuration making it much easier for new developers to start contributing to the project.

If and when we need to test against a real PostgreSQL database the developer can easily modify `QUERIDODIARIO_DATABASE_URL` to point to a real database (but this probably will be needed only in very specific cases).


jvanz:
Will you use the data stored in the database to control which item should be sent for processing pipeline?

rennerocha:
> Will you use the data stored in the database to control which item should be sent for processing pipeline?

It may be used for that too. Scrapers should only get the information and store it (downloading the files and storing in database). Any other application that wants to use our data needs to know only how to access the database and the stored files.

jvanz:
> y other application that wants to use our data needs to know only how to access the database and th

I can see the use. But as we are building the processing pipeline, we do not decided where to store this info yet. Will not be easier just upload the file to some sh3ish storage system and keep it there until we decided where we will store the pending files? AFAIK, FilePipeline already upload it automatically, if needed.

Right now, I cannot say with 100% sure that a database will be used for that.  It does not seem the more appropriate storage system to do that. However, this may or may not be true.

rennerocha:
I am 100% certain that this data will be stored in some database at some point. Storing it as raw data in a S3 bucket doesn't seem a good idea as it will be very difficult to search among hundreds of CSV files until we find something. It is not just the processing part that will need this information. If we want to make an UI to search (and we definitely should do), we will need a database to search for.

giuliocc:
This is just my idea right now of the future: the data that this project outputs would not be consumed directly (in general). The data processing part would serve the ready-to-consume data after transformations. For me, the cheapest storage would be the starting point and then change it if needed.

jvanz:
> I am 100% certain that this data will be stored in some database at some point. Storing it as raw data in a S3 bucket doesn't seem a good idea as it will be very difficult to search among hundreds of CSV files until we find something. It is not just the processing part that will need this information. If we want to make an UI to search (and we definitely should do), we will need a database to search for.

I'm not saying a database will not be necessary. I know that we will need a database at some point. This is not in discussion. What I'm saying is that the SQL *may* not be the option for the current stage. As @giuliocc  pointed out, the project will have a processing pipeline and the final storage/data used by the system may be quite different from the data right after the spider found the file. However, we just need to put the files somewhere while we decide where and how to process them. This is under discussion right now.  I'm also aware the S3 like storage is not indicated for that. But to ingest the files in the future processing pipeline is easy. The API already has ways to walk through the object(files) and we will do that just once. The idea is not search for data on the flight. We just need to keep the files there for a while until the get the processing working. In other words, temporary cheap storage.

processing/data_collection/gazette/pipelines.py
```diff
@@ -126,6 +119,52 @@ def _is_file_type(self, filepath, file_types):
         return self.get_file_type(filepath) in file_types
 
 
+class PostgreSQLPipeline:
```
jvanz, line 122: As the default DB in the repo is SQLite, why not `SQLDatabasePipeline` or something similar?

rennerocha: It makes sense. I renamed it!

jvanz (approved):

PR #286, opened 2020-10-09
Add spider for Santo André-SP

netoferraz:
The Gazzete from Santo André-SP shares your resources (pdf files) through `POST` requests. Because of that, this spider does not fit automatically on `FilesPipeline`. Despite that, we could override some of the methods (`get_media_requests`), but the entire pipeline relies on `GET` requests. Thus, I decided not to use `FilesPipeline` and make a custom configuration (`custom_settings`) for this spider.

Moreover, pagination relies on `POST` requests. Thus the automatic control to not visit duplicated URL's does not work in this scenario, so I needed to build from scratch a process to paginate the website.


processing/data_collection/gazette/spiders/sp_santo_andre.py
```diff
@@ -0,0 +1,187 @@
...
+                "ctl00$ContentPlaceHolder1$dt_inicio": self.start_date.strftime(
+                    "%d/%m/%Y"
+                ),
+                "ctl00$ContentPlaceHolder1$dt_final": self.TODAY,
+            },
+            clickdata={"name": "ctl00$ContentPlaceHolder1$PsaToolBar1$btnSelecionar"},
+            method="POST",
+        )
+
+    def _save_pdf(self, response, path: Path, item):
```
rennerocha, line 64: **suggestions**: We don't use type hints on the spiders. To keep all spiders consistent it would be better to remove it.

processing/data_collection/gazette/spiders/sp_santo_andre.py
```diff
@@ -0,0 +1,187 @@
...
+        if not isinstance(path, Path):
+            raise TypeError("O parâmetro path precisa ser uma instância de Path.")
+        path.parent.mkdir(parents=True, exist_ok=True)
+        with open(path, "wb") as f:
```
rennerocha, line 68: **issue**: We can't store the file locally. Our pipeline will store files in a S3-compatible space, so this solution is not acceptable. Probably we will need a custom FilePipeline that is able to upload the content of the PDF to a S3 bucket.

processing/data_collection/gazette/spiders/sp_santo_andre.py
```diff
@@ -0,0 +1,187 @@
...
+            str.encode(
+                f"{self.allowed_domains[0].replace('.','')}-{event_target}-{num_edicao}-{date}"
+            )
+        ).hexdigest()
+        filename = f"{filename}.pdf"
+        file_path = Path(f"{FILES_STORE}full/{filename}")
+        return file_path
+
+    def _handle_navigation(self, response, current_page: int) -> Optional[str]:
+        """Responsável por realizar a identificação da próxima página a ser navegada na paginação.
```
rennerocha, line 95: **issue**: Source code language must be English.

rennerocha (changes requested):
I know that this spider is not easy to develop. I didn't have time to run it on my machine yet, but I added a few comments of some changes we will need to make. Later I will try to run it and if I find something else I will add a new review.

processing/data_collection/gazette/spiders/sp_santo_andre.py
netoferraz: that's fine... it's very common for me mix languages while i'm coding.

processing/data_collection/gazette/spiders/sp_santo_andre.py
netoferraz: OK! I'm very susceptible to use typing annotations these days. But I understand that we need to promote a coding standard for the project.

processing/data_collection/gazette/spiders/sp_santo_andre.py
netoferraz: Perhaps, can I create another method to save these data on S3 bucket with [boto3](https://github.com/boto/boto3). What do you think? Do you have any suggestions? Or maybe we need to wait for some decisions about the infrastructure of the project? Perhaps this is not the best choice because it can freeze many on going new spiders for the project.

rennerocha: I would try at first investigate the use of a custom pipeline based on `scrapy.pipelines.files.FilesPipeline` that we use in the project specific for this spider. It would be better to keep spider's code unaware about how the files extracted are stored.

giuliocc: I agree with @rennerocha . I thought of a workaround like this, let me know what you guys think.

```python
# pipelines.py
...
from itemadapter import ItemAdapter
from scrapy.http import Request
from scrapy.pipelines.files import FilesPipeline
from scrapy.settings import Settings
...
class QueridoDiarioFilesPipeline(FilesPipeline):
    DEFAULT_FILES_REQUESTS_FIELD = 'file_requests'

    def __init__(self, *args, settings=None, **kwargs):
        super().__init__(*args, settings=settings, **kwargs)

        if isinstance(settings, dict) or settings is None:
            settings = Settings(settings)

        self.files_requests_field = settings.get(
            'FILES_REQUESTS_FIELD', self.DEFAULT_FILES_REQUESTS_FIELD
        )


    def get_media_requests(self, item, info):
        urls = ItemAdapter(item).get(self.files_urls_field, [])
        yield from (Request(u) for u in urls)
        requests = ItemAdapter(item).get(self.files_requests_field, [])
        yield from requests
```
```python
# items.py
...
class Gazette(scrapy.Item):
	...
	file_requests = scrapy.Field()
	...
```

processing/data_collection/gazette/spiders/sp_santo_andre.py
```diff
@@ -0,0 +1,187 @@
...
+        VIEW_STATE, EVENT_VALIDATION = self._get_form_params(response)
+        yield FormRequest.from_response(
+            response,
+            callback=self.parse,
+            formname="aspnetForm",
+            formdata={
+                "__EVENTTARGET": "",
+                "__EVENTARGUMENT": "",
+                "__VIEWSTATE": VIEW_STATE,
+                "__EVENTVALIDATION": EVENT_VALIDATION,
```
giuliocc, line 54: Since you are working with `FormRequest.from_response` and those `"__SOMETHING"` keys are "hidden input elements" you normally wouldn't need to manually capture its values, `from_response` already should do that for you :)

I tested it here and it worked. Could you check this too?

On the other requests, the `"__EVENTTARGET"` key was needed, though.

giuliocc (commented):
I wasn't able to run the spider at first. It showed me an error like this:

image

I was able to solve it with:

```python
from fake_useragent import UserAgent
...
	custom_settings = {
		...
		"USER_AGENT": UserAgent().random,
	}
...
```

Could you please check if that started occurring with you too or it's just me?

Made some suggestions to help you there. Nice work!

netoferraz:
> ```python
> from fake_useragent import UserAgent
> ```

Hello, @giuliocc ! Do you remember that I told you on telegram that I'm was in trouble with this site using scrapy, without splash? That's the case! I believe this site has some random behavior to detect bots. Today, I received the same message that you share with us. Besides that, congratz on this excellent turnaround to use fake user agents to handle it. That's great because we don't need to splash on it. Furthermore, after I'm downloading all PDF's available - almost 5k - this site changed his behavior, and my code stopped to work correctly.

I just ran the code, and all works as expected.

processing/data_collection/gazette/spiders/sp_santo_andre.py
netoferraz: > Since you are working with `FormRequest.from_response` and those `"__SOMETHING"` keys are "hidden input elements" you normally wouldn't need to manually capture its values, `from_response` already should do that for you :)
> 
> I tested it here and it worked. Could you check this too?
> 
> On the other requests, the `"__EVENTTARGET"` key was needed, though.

I had not aware of the `FormRequest.from_response `capabilities. I already tested your suggestion and worked fine!

processing/data_collection/gazette/spiders/sp_santo_andre.py
netoferraz: > ```python
> self.files_requests_field
> ```

I have to confess that I didn't follow you. That new field on Item, what is his purpose? The traditional field (`files_urls_field`) receives the URLs that will be used by `RequestWithItem`. Furthermore, in our scenario, we need to made POST request via `FormRequest`, right? Thereby, where our POST requests make a relationship with this new field? I had some research, and it seems that we need to send out POST to the scrapy scheduler to process the pipeline, something like that...

netoferraz:
Guys, i just already [push](a86454ec94750e94e39aacae482557b4909b1b03) some changes requested for this PR. However, the adaption for this spider use any Pipeline still is pending.

processing/data_collection/gazette/spiders/sp_santo_andre.py
giuliocc: Sorry for not clearing that up. With a pipeline like that instead of the default one we would be using a `file_requests` field on the `Gazette` item if we needed to send requests instead of `file_urls`. Building the requests instances inside the spider and sending them away inside a list (`file_requests`) through the item to be requested during pipeline processing. I felt like a new field would be better to not mix the terminology.

Oh, and the existing `QueridoDiarioFilesPipeline` is unrelated to this one. Sorry for giving it the same name and mixing those up. I've changed this one's name to make it clear.

giuliocc:
Couldn't check this before. Sorry about that :/ 

Would you prefer if I made a PR to your fork with the proposed pipeline changes?

netoferraz:
> Couldn't check this before. Sorry about that :/
> 
> @netoferraz Would you prefer if I made a PR to your fork with the proposed pipeline changes? There you could test it out and if you like it, merge with your branch.

Hey, @giuliocc ! This would be great! Push foward and i will evaluate on my fork.

rodbv:
> Couldn't check this before. Sorry about that :/
> 
> @netoferraz Would you prefer if I made a PR to your fork with the proposed pipeline changes? There you could test it out and if you like it, merge with your branch.

Hey @giuliocc I'd like to test your file pipeline suggestion on the spider I'm building for Vila Velha, can you please send a PR to my fork at https://github.com/rodbv/querido-diario? 

Cheers

giuliocc:
[Did it](https://github.com/rodbv/querido-diario/pull/1) @rodbv . In the meanwhile, I'll work with @netoferraz in setting up his project. Hopefully, if everything works out, we'll be able to use the modified pipeline soon.

PR #88, opened 2018-06-09
spider: add spider for São Gonçalo, RJ

chicocvenancio:

PR #292, opened 2020-10-10
Add spider for São Gonçalo, RJ

vitorbaptista:
I merged #88 by @chicocvenancio with the current main branch and fixed some updates needed in the code. This is ready to be merged :+1:


processing/data_collection/gazette/spiders/rj_sao_goncalo.py
```diff
@@ -0,0 +1,35 @@
...
+    start_urls = ["https://servicos.pmsg.rj.gov.br/diario_oficial.php"]
+
+    def parse(self, response):
+        date = datetime.date.today()
+        start_date = self.start_date
+        while date >= start_date:
+            is_weekend = date.weekday() >= 5
+            if is_weekend:
+                date = date - datetime.timedelta(days=1)
+                continue
```
giuliocc, line 23: I found this interesting, took it with a grain of salt at first but after testing this seems really nice :)

giuliocc (commented):
Working fine :) Let's just consider that 404's are not a good measure of spider health in this case hehehehe.

processing/data_collection/gazette/spiders/rj_sao_goncalo.py
```diff
@@ -0,0 +1,35 @@
...
+
+            url = self.DOWNLOAD_GAZETTE_URL.format(date.strftime("%Y_%m_%d"))
+            # TODO: How to figure out is_extra_edition? Doesn't seem possible
+            # without extracting the text.
```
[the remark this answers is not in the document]
vitorbaptista: @giuliocc I like these TODOs in code, as a "lightweight issue". I find useful to be able to search for TODOs and see what's pending. However, I'm happy to remove it. What do you think?

processing/data_collection/gazette/spiders/rj_sao_goncalo.py
```diff
@@ -0,0 +1,35 @@
...
+    name = "rj_sao_goncalo"
+    allowed_domains = ["servicos.pmsg.rj.gov.br"]
+    start_urls = ["https://servicos.pmsg.rj.gov.br/diario_oficial.php"]
+
+    def parse(self, response):
+        date = datetime.date.today()
+        start_date = self.start_date
+        while date >= start_date:
+            is_weekend = date.weekday() >= 5
+            if is_weekend:
```
rennerocha, line 21: **issue (blocking)** We **do** have gazettes on weekends. I ran this spider replacing this line with `if not weekend` just to get items only for weekends. The majority doesn't have a gazette, however I found 527 gazettes on weekends. For example 11/03/2000 is a Saturday and we have a gazette.

Nevertheless I found something interesting there. We can get the gazettes straight from the generated URLs, but if we search for this date in the website, we don't have any response. It is a bug on their site.

rennerocha (changes requested):
@vitorbaptista thanks for the PR! 
I found a blocking issue there that make the spider lose some gazettes. Please take a look over it. Otherwise the spider is fine and it was great that you already used `start_date` as a way to reduce the amount of unneeded requests (:tada:).

processing/data_collection/gazette/spiders/rj_sao_goncalo.py
giuliocc: You're right, I didn't test this for the entire timespan. It's good that you caught it, thanks ;)

processing/data_collection/gazette/spiders/rj_sao_goncalo.py
giuliocc: Tested this from 2018 to now and as @rennerocha pointed out it does, in fact, not work because of older dates :/

rennerocha:
Closed as stale.

Issue #267, opened 2020-10-03
Monitor if returned items follows expected field type


rennerocha:
With Spider monitoring in place (#248), we will be able to start monitoring if the items that we are returning matches a specific schema ([as it is explained here](https://spidermon.readthedocs.io/en/latest/item-validation.html#with-schematics)). This will avoid problems like https://github.com/okfn-brasil/querido-diario/pull/263/files .

Tasks needed:
- [ ] If #248 is not implemented yet, enable Spidermon into the project. If it is implemented, just need to do the next task;
- [ ] Add monitor that check if all returned fields matches a well defined schema

andersonberg:
I can work on this issue. I can start creating the schema while https://github.com/okfn-brasil/querido-diario/issues/248 is not completed

rennerocha:
Closed by https://github.com/okfn-brasil/querido-diario/pull/304 .

PR #302, opened 2020-10-11
Spidermon schema

andersonberg:
This PR will handle https://github.com/okfn-brasil/querido-diario/issues/267 also adding and enabling Spidermon.


andersonberg:
While working on this PR, I found a bug in the Cuiaba Spider that was returning `None` for the edition date. Since it's a small change I find it better to include in this PR

processing/data_collection/gazette/schema.json
```diff
@@ -0,0 +1,58 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "type": "object",
+    "properties": {
+        "date": {
+            "description": "Date of the gazzete",
+            "type": "string"
```
rennerocha, line 7: **issue**: Gazette date is a well defined string with YYYY-MM-DD. It would be a good idea to match this pattern when checking this field.

processing/data_collection/gazette/settings.py
```diff
@@ -10,3 +10,13 @@
 }
 FILES_STORE = "/mnt/data/"
 QUERIDODIARIO_EXTRACT_TEXT_FROM_FILE = True
+
+EXTENSIONS = {
+    "spidermon.contrib.scrapy.extensions.Spidermon": 500,
+}
+SPIDERMON_ENABLED = True
+ITEM_PIPELINES = {
```
rennerocha, line 18: **issue (blocking)**: We already have a list of ITEM_PIPELINES defined. This change will disable all other pipelines.

processing/data_collection/gazette/spiders/mt_cuiaba.py
```diff
@@ -42,7 +42,7 @@ def parse(self, response):
             edition_date = edition["publication_date"]
 
             yield Gazette(
-                date=parse(edition_date, languages=["pt"]),
+                date=parse(edition_date),
```
rennerocha, line 45: **suggestion**: This change is not related to Spidermon, so it shouldn't be here. Also when parsing a date using dateparser, it is good practice to pass the language to make the parsing more reliable.

rennerocha (changes requested):
@andersonberg thanks for the PR! I added a few comments but we have few changes before getting this merged.

processing/data_collection/gazette/schema.json
```diff
@@ -0,0 +1,58 @@
...
+        "is_extra_edition": {
+            "description": "Determines if the edition of the gazzete is extra or not",
+            "type": "boolean"
+        },
+        "territory_id": {
+            "type": "string"
+        },
+        "power": {
+            "type": "string",
+            "examples": ["executive", "legislative"]
```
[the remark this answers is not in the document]
andersonberg: this is only a list with examples, but I can add a `enum` with possible values

processing/data_collection/gazette/schema.json
andersonberg: that's a good idea, I thought the date were a "datetime" object

processing/data_collection/gazette/spiders/mt_cuiaba.py
andersonberg: ok I'll add a new PR to fix this. Maybe change the language, because using "pt", dateparser were returning `None`

rennerocha (approved):
@andersonberg 
I did a few small changes to make Spidermon work correctly.

andersonberg:
Thank you @rennerocha! Looks good to me

PR #305, opened 2020-10-12
Removes from the spiders the autogenerated gazette fields

tcurvelo:
Since #253, spiders don't need to set `scraped_at` and `territory_id` at gazettes creation, saving us a few keystrokes. 
So I decided to do some housecleaning and removed them from all spiders :sleepy:


giuliocc (approved):
LGTM. Did you run `autoflake` too? :)

jvanz (approved):
LGTM. But I would say that you do not need to put the flake8 command used in the commit message. Just a description what you did is sufficient. If you rewrite the commit message it would be great.

PR #306, opened 2020-10-12
File processed flag

jvanz:
Adds a new column in the Gazette model. This column should be used by the processing pipeline to find which gazettes is pending to be ingested by the system. This is not the best solution to this, a queue system would be a better approach for instance. But for the current state of the project this will be enough.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,3 +123,4 @@ Sessionx.vim
 *~
 tags
 [._]*.un~
+querido-diario.db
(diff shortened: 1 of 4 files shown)


giuliocc (approved):
Since this doesn't disrupt anything and you find it necessary for now: LGTM.

Both the build and the `make sql` have been tested and I found no issues.

Kudos on correcting those ;)

rennerocha:
> Adds a new column in the Gazette model. This column should be used by the processing pipeline to find which gazettes is pending to be ingested by the system. This is not the best solution to this, a queue system would be a better approach for instance. But for the current state of the project this will be enough.

This field existed before and I removed it :smile: . But I think it will be good to keep it at this moment so you can verify more easily which gazettes were processed in the pipeline you are developing. LGTM!

rennerocha (approved):

PR #310, opened 2020-10-15
Remove text extraction from pipeline

rennerocha:
Processing of content of downloaded files will be done in a separated pipeline, so we don't need it in this part of the process.

### DIFF
--- a/processing/Dockerfile
+++ b/processing/Dockerfile
@@ -6,13 +6,9 @@ WORKDIR /mnt/code
 
 COPY . /mnt/code
 
-# install Apache Tika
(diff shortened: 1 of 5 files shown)


jvanz:
LGTM. But I would like to take a feedback from @aassumpcao. Do you want to keep the pipeline step there until we get the infra running? Or are you okay to have a patch applied locally your system to do that for a while (if needed) ?

jvanz (approved):

giuliocc (approved):
LGTM²

rennerocha:
@jvanz @aassumpcao 
We can reintroduce this pipeline if needed in future. However I believe from what we've been talking, it will be very unlikely that this will be part of this step of the processing.

Issue #313, opened 2020-10-16
Add possibility to store information of more than one file per-gazette


rennerocha:
In a [gazette item](https://github.com/okfn-brasil/querido-diario/blob/main/processing/data_collection/gazette/database/models.py#L31), we consider that each gazette is an unique file, so we store only the information of [one file](https://github.com/okfn-brasil/querido-diario/blob/main/processing/data_collection/gazette/database/models.py#L39).

This is true most of the times, but it is possible that the same gazette is splitted in more than one file like we have in [Piracicaba-SP](https://github.com/okfn-brasil/querido-diario/blob/main/processing/data_collection/gazette/spiders/sp_piracicaba.py#L33) that for some gazettes, it contains two PDF files (each file has a maximum of 50 pages).

So we may lose important information that will be necessary when we process the downloaded files and extract their content.

giuliocc:
:/ 

What about a `order_in_sequence` integer field? Just spilling thoughts right now.

If we can get the order from the system (pages, chunks, etc.), great. When processing we can concatenate in a ascending order that the spider extracted. 

If we can't. Well, there's not much to do in the spider and we try to solve it when processing. Maybe case by case.

rennerocha:
This is not really a problem in the Spider. We already pass a list of URLs in `file_urls` field and Scrapy will download every file from there. Usually we have only one file, but it seems that we have some few cities that split the gazettes in more than one file.

The problem will appear when accessing the database table. This will be required by the processing step, when we extract the text from the downloaded files. In my opinion, the way to solve it is to change the database structure to allow us to store the information of as many downloaded files we have.

Something like:
```
class Gazette(DeclarativeBase):
    __tablename__ = "gazettes"
    id = Column(Integer, primary_key=True)

class GazetteFile(DeclarativeBase):
    __tablename__ = "gazettefiles"
    sequence = Column(Integer)  # This acts like order_in_sequence you suggested
    checksum = Column(String)
    path = Column(String)
    url = Column(String)
    gazette = relationship("Gazette", back_populates="files")
```

And then update `SQLDatabasePipeline` to deal with this new structure (considering that we should not add duplicated entries if the checksum is the same), etc ...

giuliocc:
That's what I was thinking :)

jvanz:
[...]

PR #316, opened 2020-10-16
Fix ExtractTextPipeline and SQLDatabasePipeline

phrfpeixoto:
This is a simple fix for #256. Reopening after #257 

> Thanks for your PR @phrfpeixoto ! But I`ll not accept it. We are in the process of moving off the data manipulation code from this repository. So, this code will be obsolete within days. Sorry... =/
>
> If you interesting in check the future commits on that. Keep eyes on this repo: https://github.com/okfn-brasil/querido-diario-data-processing

I see your point, but I really don't think this is a massive change in the code, and having it accepted would not pose any challenge in this "migration" to a new repository.

On the other hand it could improve development process on spiders in here. This is a change that I often have to do whenever I need to run a new spider and avoid downloading files.

Furthermore, the pipeline does have a bug. If you use any way to avoid downloading files, it will break. The only way to avoid it is to disable the pipeline: Either supressing the files_url on the scraped items, or disabling the FILES_STORE directory

SPIDER="ap_macapa -s FILES_STORE='' -a start_date=2020-09-01 -a end_date=2020-09-30 -o /mnt/data/items.json" make run_spider

This still breaks. It would have to be:

SPIDER="ap_macapa -s QUERIDODIARIO_EXTRACT_TEXT_FROM_FILE=0 -s FILES_STORE='' -a start_date=2020-09-01 -a end_date=2020-09-30 -o /mnt/data/items.json" make run_spider

Doesn't seem right to me having to disable the pipeline to work-around a bug that can be fixed with 2 lines of code.

### DIFF
--- a/processing/data_collection/gazette/pipelines.py
+++ b/processing/data_collection/gazette/pipelines.py
@@ -50,6 +50,9 @@ class ExtractTextPipeline:
         if not extract_text_from_file:
             return item
 
+        if "files" not in item or not item["files"]:
+            return item
+
         if self.is_doc(item["files"][0]["path"]):
             item["source_text"] = self.doc_source_text(item)
         elif self.is_pdf(item["files"][0]["path"]):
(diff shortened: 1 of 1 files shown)


phrfpeixoto:
@giuliocc I can't add you as a reviewer, but could you please consider this?

rennerocha:
Hello @phrfpeixoto !

Thanks for the PR! It really solves the problem we may face during development and would make developers life much easier.

However, we are removing `ExtractTextPipeline` in PR https://github.com/okfn-brasil/querido-diario/pull/310 and PR https://github.com/okfn-brasil/querido-diario/pull/315 solved a serious bug #313 that changed the way we process files in `SQLDatabasePipeline`. So the changes in this PR will not be needed as they are changing parts of the code that were removed or changed in a very different way.

phrfpeixoto:
Thanks for clarifying.

PR #318, opened 2020-10-18
Add Vila Velha/ES

### LINKED ISSUE #319: Add spider for Vila Velha/ES
There is a city of Vila Velha with a green tick on CITIES.md but the spider for `es_associacao_de_municipios` does not include Vila Velha, as far as I could see (the second page of this gazette has the list of cities included and Vila Velha is not there). So I will create a specific spider for Vila Velha.

Their gazette repository is at https://www.vilavelha.es.gov.br/diariooficial/ConsultaDiario.aspx.

rodbv:
Spider for Vila Velha/ES (initial work for #319)

Note: There is a city of Vila Velha with a green tick on CITIES.md but the spider for `es_associacao_de_municipios` does not include Vila Velha, as far as I could see (the second page of this gazette has the list of cities included and Vila Velha is not there). So I created a specific spider for Vila Velha.

Their gazette repository is at https://www.vilavelha.es.gov.br/diariooficial/ConsultaDiario.aspx

The biggest challenge for this gazette is that there are no `GET` routes for the gazette PDFs, the page is built using ASP.NET and the gazettes are downloaded via a `POST` method behind a link with a javascript function call, e.g. 

```html
<a href="javascript:__doPostBack('ctl00$cpConteudo$gvDocumentos$ctl02$lbDownloadPDF','')">
```

The only solution I could come up with is to save the file while the spider finds those links, but I believe this is not the right way. 

I can't return the fileUrl on the gazette object, it would have to be some sort of callback function, or I'd have to save the PDF in some sort of temporary place and then give this filesystem path as fileUrl, so instead of running a "GET [fileUrl]" it would just move a file in the filesystem, and then assign the correct filename etc. But both alternatives go beyond the code for the spider, unless scrapy already supports this kind of scenarios. 

Any help is appreciated!

Edit: this PR closes #319


giuliocc:
@rodbv this problem is exactly the same as in #286 (ASP.NET FTW!!! \s). We are currently trying to solve it in a way that integrates better into the project :)

rodbv:
Thanks @giuliocc, I'm having a look. At least on Vila Velha's case there are no pagination and user agent issues, just the POST challenge. I'm glad to help to evolve the pipeline code as well

rodbv:
hey @giuliocc I merged your file pipeline changes, and noticed that the spider for Vila Velha was adjusted already, thanks! :) 

As you probably also saw in your branch as well, it worked beautifully!

One thing I noticed is that the files downloaded for Vila Velha didn't get the .pdf extension, while other spiders I've tried add this extension. Not sure if that's an important issue though.

Thanks a lot

processing/data_collection/gazette/spiders/es_vila_velha.py
```diff
@@ -0,0 +1,56 @@
...
+    name = "es_vila_velha"
+    allowed_domains = ["www.vilavelha.es.gov.br"]
+
+    TERRITORY_ID = "3205200"
+    GAZETTE_URL_CSS = "td:last-child a::attr(href)"
+    GAZETTE_DATE_CSS = "td:nth-child(2) span b::text"
+    GAZETTE_ISSUE_CSS = "td:nth-child(3) span b::text"
+    JAVASCRIPT_POSTBACK_REGEX = r"javascript:__doPostBack\('(.*)',''\)"
+
+    # this is the first date which has gazettes available (edition number 1)
```
giuliocc, line 20: I don't think this comment is necessary since it is a project pattern. Do you have any thoughts about this?

processing/data_collection/gazette/spiders/es_vila_velha.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def start_requests(self):
+        date_param = self.start_date.strftime("%d/%m/%Y")
+        base_url = "https://www.vilavelha.es.gov.br/diariooficial/ConsultaDiario.aspx"
+        gazettes_url = f"{base_url}?dataInicial={date_param}"
+        yield Request(gazettes_url)
+
+    def parse(self, response):
+        for element in response.css("#ctl00_cpConteudo_gvDocumentos tr"):
+            if element.css("th").extract():
```
giuliocc, line 31: Could you attribute this to a variable with a meaningful name so it gets easier to understand why this type of element is needed to be skipped?

processing/data_collection/gazette/spiders/es_vila_velha.py
```diff
@@ -0,0 +1,56 @@
...
+                continue
+
+            date = element.css(self.GAZETTE_DATE_CSS).extract_first()
+            date = dateparser.parse(date, languages=["pt"]).date()
+            url = element.css(self.GAZETTE_URL_CSS)
+            gazette_issue = element.css(self.GAZETTE_ISSUE_CSS).extract_first()
+            is_extra = "EXTRA" in gazette_issue
+            edition_number = gazette_issue.split(" ")[0]
+            event_target = url.re(self.JAVASCRIPT_POSTBACK_REGEX).pop()
```
giuliocc, line 40: Since url isn't used, do you think this would be better?
```suggestion
            event_target = element.css(self.GAZETTE_URL_CSS).re_first(self.JAVASCRIPT_POSTBACK_REGEX)
            gazette_issue = element.css(self.GAZETTE_ISSUE_CSS).extract_first()
            is_extra = "EXTRA" in gazette_issue
            edition_number = gazette_issue.split(" ")[0]
```

processing/data_collection/gazette/spiders/es_vila_velha.py
```diff
@@ -0,0 +1,56 @@
...
+
+            document_request = FormRequest.from_response(
+                response,
+                formdata={"__EVENTARGUMENT": "", "__EVENTTARGET": event_target},
```
giuliocc, line 44: In my tests, `"__EVENTARGUMENT"` wasn't necessary, could you confirm that?
```suggestion
                formdata={"__EVENTTARGET": event_target},
```

processing/data_collection/gazette/spiders/es_vila_velha.py
```diff
@@ -0,0 +1,56 @@
...
+                dont_click=True,
+                dont_filter=True,
```
giuliocc, line 46: In my tests, `dont_filter` and `dont_click` weren't necessary.

giuliocc (changes requested):
Nice! Glad you liked it ;)

Commit f89e374 can be removed since it is a duplicate of 314f405.

Also, I think this PR can already be opened, right?

I didn't test the entire timespan while downloading the files because of disk space, but the last year is OK in this regard.

giuliocc:
@rennerocha could you review it? I'm pretty biased right now :)

processing/data_collection/gazette/spiders/es_vila_velha.py
rodbv: I just wanted to make clear that there was no point setting the default date to anything before 2016-07-01, but on the other hand I noticed that if we try dates like 2010-01-01 it gets _much_ slower to get a response back, probably due to the way the database query works over there in Vila Velha. But maybe that's what should have been in the comment if I want to keep some :)

processing/data_collection/gazette/spiders/es_vila_velha.py
rodbv: True! thanks

rodbv:
Thanks for the review @giuliocc 
I got one newcomer question: should my PR have only the file for the Vila Velha spider (only that single file)? right now it also includes the changes you've done  on the pipeline, since I've incorporated your PR in my branch

[https://github.com/okfn-brasil/querido-diario/pull/318/files](https://github.com/okfn-brasil/querido-diario/pull/318/files)

If it should just be the spider file itself maybe I should create  a new branch with just that and change the source branch?

processing/data_collection/gazette/spiders/es_vila_velha.py
giuliocc: Hm... Understood :) Yeah, but since dates before the spider's `start_date` shouldn't be run, we shouldn't have this kind of problem and that kind of comment wouldn't be necessary too.

processing/data_collection/gazette/spiders/es_vila_velha.py
```diff
@@ -0,0 +1,54 @@
...
+        base_url = "https://www.vilavelha.es.gov.br/diariooficial/ConsultaDiario.aspx"
+        gazettes_url = f"{base_url}?dataInicial={date_param}"
+        yield Request(gazettes_url)
+
+    def parse(self, response):
+        table_head = "th"
+
+        for element in response.css("#ctl00_cpConteudo_gvDocumentos tr"):
+            if element.css(table_head).extract():
+                continue
```
giuliocc, line 33: What do you think about this?
```suggestion
        for element in response.css("#ctl00_cpConteudo_gvDocumentos tr"):
            is_header = element.css("th").extract() != []
            if is_header:
            	continue
```

giuliocc:
> I got one newcomer question: should my PR have only the file for the Vila Velha spider (only that single file)? right now it also includes the changes you've done on the pipeline, since I've incorporated your PR in my branch
> 
> [#318 (files)](https://github.com/okfn-brasil/querido-diario/pull/318/files)
> 
> If it should just be the spider file itself maybe I should create a new branch with just that and change the source branch?

I don't think it should be necessary. These are not big changes. And since the first use case is this spider, it makes sense to me to bundle them together.

processing/data_collection/gazette/spiders/es_vila_velha.py
rodbv: Code review improvements

rodbv: I like it 👍

giuliocc (approved):

rodbv:
PR #348 will replace this one, due to the several changes in the project structure.

Issue #320, opened 2020-10-18
Should we add instructions to create a virtual environment before running "make setup"?


rodbv:
The instructions on README.md suggest to run `make setup` without a virtual environment, which will run `pip3 install` on the user's machine, not isolated in a virtual environment, and is prone to dependency conflicts etc.

We should leverage the benefits of virtual environments, and ask the user to create and activate one before they run `make setup`.

If this makes sense I'm glad to send this PR with instructions for Mac/Linux and Windows users.

rennerocha:
@rodbv hopefully this was fixed after https://github.com/okfn-brasil/querido-diario/pull/323 were merged.
I will close this issue. Please open it again if you think we still need to change the documentation.

PR #323, opened 2020-10-19
Remove Docker dependency for development

rennerocha:
After removing text extraction (https://github.com/okfn-brasil/querido-diario/pull/310) and database dependency (https://github.com/okfn-brasil/querido-diario/commit/0447f2b2f4e342346c50bbe4a0c0abd512584ce8) from the project, requiring Docker to install and use the project seems to be a barrier for new developers, especially considering that we are not using anything that really requires a Docker container to be running in developers machine.

In my opinion, a good old virtualenv, that can be created using built-in `[venv](https://docs.python.org/3/library/venv.html)` module is enough to create a development environment suitable for the majority of developers.

On this PR I am suggesting to remove any reference to Docker to make developers life easier.

When (and if) we need to deploy this project in another environment on production, we can chose if Docker is the best solution, but these configurations can be separated and not affecting people that only wants to contribute with the spiders.


rodbv:
This would also close issue #320 which I opened yesterday, to add `venv` instructions on README.md

jeanmask:
I agree, decouple database and text extraction pipelines can be a good idea to help new interested collaborators.

giuliocc:
I think it does make sense to remove it. Not sure if we will ever go back to using it, but if we do it shouldn't be a problem anyways.

If I may suggest a few things:
- Move the `data_processing` directory to the project root
- Create a `requirements` folder with `requirements.[in,txt]` and `dev.[in,txt]`
- Change `Makefile` to reflect commands that usually were made using Docker

jvanz (approved):

rennerocha:
Replaced in favor of https://github.com/okfn-brasil/querido-diario/pull/341 (that solves some weird conflicts).

PR #327, opened 2020-10-23
Fix power field to the values expected by spidermon

jvanz:
After the spidermon configuration some spider stop working due a invalid
value for the power item field. This commit replace all these invalid
values for the value allowed by the spidermon schema


giovanisleite (approved):

giuliocc (changes requested):
`rs_porto_alegre` and `pr_cascavel` can yield `legislature` power. The rest is good :)

rennerocha (approved):

PR #328, opened 2020-10-25
Misleading error message when write in the database fails.

jvanz:
The integrity error raised by SQLAlchemy is launched not only when the
item is already in the database. For instance, it can also be launched
when the territories table is missing some territory ID. Thus, this
commit updates the error message and add the exception in the log
message. Thus, the users will be able to detect which kind of error
they are facing.


giuliocc (approved):
Seems good. A simple rebase got it working here

Issue #332, opened 2020-10-27
Caucaia spider


Jefersonalves:
Add spider for Caucaia/CE

PR #335, opened 2020-10-27
Add spider for Caucaia/CE

Jefersonalves:
This PR solves #332. This spider already makes fewer requests based on the start date


processing/data_collection/gazette/spiders/ce_caucaia.py
```diff
@@ -0,0 +1,77 @@
...
+        5: "Maio",
+        6: "Junho",
+        7: "Julho",
+        8: "Agosto",
+        9: "Setembro",
+        10: "Outubro",
+        11: "Novembro",
+        12: "Dezembro",
+    }
+
```
jvanz, line 38: I think you can use the standard for that. Take a look in the calendar module.

jvanz (changes requested):

jvanz (approved):

PR #337, opened 2020-10-29
Initialize database with territories information

rennerocha:
To be able to insert data into `gazettes` table, we need to have the corresponding record in `territories` table, as we have a relationship between these two tables. We have a list of all territories available as a CSV file. With this PR, when we start a spider, it will check and populate `territories` table if needed. This is not failing with SQLite, but it will fail when we use PostgreSQL database.


jvanz (approved):

jvanz:
@rennerocha The code in this PR is also present in the  #338. Am I right? Do you prefer merge this one and the #339 or only the #338 ?

jvanz:
If we can merge this one. Can we close #338 ?

rennerocha:
Merge this one. I did it first.

PR #341, opened 2020-11-02
Remove docker fixing conflicts

rennerocha:
This PR will replace https://github.com/okfn-brasil/querido-diario/pull/323 (because some weird conflicts that I was unable to solve).
It would be great if we merge this before any other PR, or else we can have conflicts again ...


jvanz:
Ops... I think that I see this one too late. =/

PR #357, opened 2020-11-07
Configure pre-commit on project

rennerocha:
Add set of pre-commit hooks to ensure that commited files are correctly formatted.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,7 @@ Project uses [Black](https://github.com/psf/black) as an automated tool to forma
 formatted according these tools.
 
 If you followed the setup instructions, installing pre-commit hooks, it is possible that you will never
-need to run these tools manually, as they will be execute before eacho commit. However, if you want
+need to run these tools manually, as they will be execute before each commit. However, if you want
 to run them in all files in the project, you have `make format` command that will call these tools.
 
 ## Guidelines to maintainers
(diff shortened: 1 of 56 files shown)


giuliocc (changes requested):
Tested it and it is running fine. One thing that I thought could be done is to make Travis run `pre-commit run --all-files` what do you think?

A suggested improvement for `CONTRIBUTING.md` on section `Automated code formatting`:

---
## Automated code formatting

The project uses tools like [Black](https://github.com/psf/black) and [isort](https://github.com/PyCQA/isort) to format code automatically. Although, if you followed the setup correctly, you wouldn't need to run these tools, because they are run automatically with [pre-commit](https://pre-commit.com/) when you are about to make a commit.

It is important to note that the CI will fail if you commit Python code that is not in accordance with Black code style.
---

CONTRIBUTING.md
```diff
@@ -42,18 +42,13 @@ SPIDER=rs_porto_alegre make run_spider
...
-```sh
-$ pip install black
-$ black .
-$ pip install -r requirements.txt
-$ make check
-```
-
-It is important to note that the CI will fail if you commit Python code that is not in accordance with Black code style.
+If you followed the setup instructions, installing pre-commit hooks, it is possible that you will never
+need to run these tools manually, as they will be execute before eacho commit. However, if you want
```
giuliocc, line 50:
```suggestion
need to run these tools manually, as they will be execute before each commit. However, if you want
```

giuliocc (changes requested):

giuliocc (approved):

PR #348, opened 2020-11-03
Spider for Vila Velha/ES

rodbv:
Closes #319 

This PR replaces #318 which will be closed,  due to the several changes on the project structure.

Code to support `file_requests` in the pipeline was provided by @giuliocc 👏

PR #359, opened 2020-11-07
Review Belém-PA spider

rennerocha:
- Improve method to obtain number of items that exists in the period
  desired without relying on blocking requests
- Code formatting
- Add new field edition_number
- Filter gazettes scraped considering start_date argument

### DIFF
--- a/data_collection/gazette/pipelines.py
+++ b/data_collection/gazette/pipelines.py
@@ -103,3 +103,10 @@ class QueridoDiarioFilesPipeline(FilesPipeline):
         datestr = item["date"].strftime("%Y-%m-%d")
         filename = Path(filepath).name
         return str(Path(item["territory_id"], datestr, filename))
+
+    def get_media_requests(self, item, info):
+        urls = ItemAdapter(item).get(self.files_urls_field, [])
+
+        download_file_headers = getattr(info.spider, "download_file_headers", {})
+
+        return [Request(u, headers=download_file_headers) for u in urls]
--- a/data_collection/gazette/spiders/pa_belem.py
+++ b/data_collection/gazette/spiders/pa_belem.py
@@ -2,23 +2,9 @@ import datetime
 from urllib.parse import urlencode
 
 import scrapy
-from itemadapter import ItemAdapter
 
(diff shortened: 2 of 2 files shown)


giuliocc (changes requested):
The files that were downloaded here had something like this as content:

```
{"id":14103,"data_publicacao":"2020-10-20T03:00:00.000+0000","attr_stream_name":["dom_20-10-20.pdf"],"attr_stream_size":["416959"]}
```

Is there a missing step?

The previous steps and date filtering are working like a charm, though!

rennerocha:
> The files that were downloaded here had something like this as content:
> 
> ```
> {"id":14103,"data_publicacao":"2020-10-20T03:00:00.000+0000","attr_stream_name":["dom_20-10-20.pdf"],"attr_stream_size":["416959"]}
> ```
> 
> Is there a missing step?
> 
> The previous steps and date filtering are working like a charm, though!

We need to add a header to the request, or else we will get this JSON content. I created a custom pipeline for this spider. This solves the problem.

rennerocha:
@giuliocc I will change this spider again. It is more complicated than it should be. Wait for an update in the next days. :-)

giuliocc:
> We need to add a header to the request, or else we will get this JSON content. I created a custom pipeline for this spider. This solves the problem.

Huum. Now the previous `DEFAULT_REQUEST_HEADERS` makes sense :) Do you prefer the pipeline implementation over the custom setting?

> @giuliocc I will change this spider again. It is more complicated than it should be. Wait for an update in the next days. :-)

Sure :+1: If my comment above doesn't make sense with what you discovered now just ignore it ;)

rennerocha:
@giuliocc 

Creating a custom pipeline works, but the procedure to replace the existing file pipeline to the new one is a bit complicated (I didn't like to override the `update_settings` method). I think a more straightforward solution is to create a new custom property in the spider with the desired headers and update our custom file pipeline.

This is a more general solution and can be used by other spiders if needed.

giuliocc (approved):
That's nice :)

It has some overlap with #348 but I think this should be approved and we leave discussions about it for that PR.

Issue #370, opened 2020-11-09
Add spider for São João de Meriti - RJ


fariaseduv:
Hi,

According to the CITIES.md there is no spider, issue or PR for São João de Meriti - RJ.

The Official Gazette could be found here: https://transparencia.meriti.rj.gov.br/?serv=1045. Their first data is from September/2011 and they always use PDFs.

I will be working in the spider for the city.

Cheers,

Eduardo Farias

fariaseduv:
Rodando pra todo o período Set-2011 -> Nov-2020:

[scrapy.statscollectors] INFO: Dumping Scrapy stats:
{'downloader/exception_count': 4,
 'downloader/exception_type_count/twisted.internet.error.TimeoutError': 3,
 'downloader/exception_type_count/twisted.web._newclient.ResponseFailed': 1,
 'downloader/request_bytes': 571327,
 'downloader/request_count': 2090,
 'downloader/request_method_count/GET': 2090,
 'downloader/response_bytes': 4000693339,
 'downloader/response_count': 2086,
 'downloader/response_status_count/200': 2054,
 'downloader/response_status_count/500': 32,
 'elapsed_time_seconds': 1168.597895,
 'file_count': 1951,
 'file_status_count/downloaded': 1951,
 'finish_reason': 'finished',
 'finish_time': datetime.datetime(2020, 11, 10, 5, 14, 35, 461443),
 'item_dropped_count': 13,
 'item_dropped_reasons_count/DropItem': 13,
 'item_scraped_count': 1953,
 'log_count/DEBUG': 5994,
 'log_count/ERROR': 12,
 'log_count/INFO': 54,
 'log_count/WARNING': 36,
 'response_received_count': 2064,
 'retry/count': 26,
 'retry/max_reached': 10,
 'retry/reason_count/500 Internal Server Error': 22,
 'retry/reason_count/twisted.internet.error.TimeoutError': 3,
 'retry/reason_count/twisted.web._newclient.ResponseFailed': 1,
 'scheduler/dequeued': 102,
 'scheduler/dequeued/memory': 102,
 'scheduler/enqueued': 102,
 'scheduler/enqueued/memory': 102,
 'spidermon/validation/fields': 14386,
 'spidermon/validation/fields/errors': 13,
 'spidermon/validation/fields/errors/field_too_short': 13,
 'spidermon/validation/fields/errors/field_too_short/files': 13,
 'spidermon/validation/items': 1966,
[...]

PR #376, opened 2020-11-11
Spider São João de Meriti - RJ

fariaseduv:
Issue https://github.com/okfn-brasil/querido-diario/issues/370

### DIFF
--- a/data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
+++ b/data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
@@ -92,8 +92,6 @@ class SpSaoJoaoDeMeritiSpider(BaseGazetteSpider):
 
         return gazettes
 
-    # FROM 2018 ONWARDS:
-
     def parse_gazette_from_2018(self, gazette_entry):
         edition_number = int(re.search(r"\d+", gazette_entry["ANEXO"]).group())
         is_extra_edition = "EXTRA" in gazette_entry["ANEXO"]
--- a/data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
+++ b/data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
@@ -55,8 +55,6 @@ class SpSaoJoaoDeMeritiSpider(BaseGazetteSpider):
             months.append(self.end_date)
         return months
 
-    # BEFORE 2018:
-
     def get_day_fom_pdf_link(self, pdf_link):
         day_regex = r"[/-](\d{1,2})[^\d/][^/]*.pdf"
         day = int(re.search(day_regex, pdf_link, re.IGNORECASE).group(1))
--- a/data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
+++ b/data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
@@ -23,24 +23,28 @@ class SpSaoJoaoDeMeritiSpider(BaseGazetteSpider):
     end_date = datetime.date.today()
 
     def start_requests(self):
-        janeiro_2018 = datetime.date(2018, 1, 1)
-        for month in self.months_interval():
-            if month < janeiro_2018:
+        january_2018 = (2018, 1)
+        for year, month in self.months_interval():
+            if (year, month) < january_2018:
                 yield scrapy.Request(
-                    f"{self.URL_BEFORE_2018}{month.year}{month.month:02}.php",
-                    callback = self.parse_month_before_2018
+                    f"{self.URL_BEFORE_2018}{year}{month:02}.php",
+                    callback=self.parse_month_before_2018,
+                    cb_kwargs={"year": year, "month": month},
                 )
             else:
                 yield scrapy.Request(
-                    f"{self.URL_FROM_2018}{month.month}/{month.year}",
-                    callback = self.parse_month_from_2018
+                    f"{self.URL_FROM_2018}{month}/{year}",
+                    callback=self.parse_month_from_2018,
                 )
 
     def months_interval(self):
-        month_rule = rrule(freq=MONTHLY, dtstart=self.start_date, until=self.end_date)
-        months = [month.date() for month in month_rule]
-        if months and months[-1].month != self.end_date.month:
-            months.append(self.end_date)
+        month_rule = rrule(
+            freq=MONTHLY,
+            dtstart=datetime.date(self.start_date.year, self.start_date.month, 1),
+            until=self.end_date,
+        )
+        months = [(month.year, month.month) for month in month_rule]
+
         return months
 
     def get_day_fom_pdf_link(self, pdf_link):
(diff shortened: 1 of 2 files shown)


data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,125 @@
...
+    URL_FROM_2018 = (
+        "https://transparencia.meriti.rj.gov.br/diario_oficial_get.php?mesano="
+    )
+
+    start_date = datetime.date(2011, 9, 30)
+    end_date = datetime.date.today()
+
+    # start_date = datetime.date(2017, 12, 25)
+    # end_date = datetime.date(2018, 1, 5)
+
```
jvanz, line 27:
```suggestion
```

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,125 @@
...
+    def months_interval(self):
+        month_rule = rrule(freq=MONTHLY, dtstart=self.start_date, until=self.end_date)
+        months = [month.date() for month in month_rule]
+        if months and months[-1].month != self.end_date.month:
+            months.append(self.end_date)
+        return months
+
+    # BEFORE 2018:
+
```
jvanz, line 59:
```suggestion
```

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,125 @@
...
+        for pdf_link in self.get_pdf_links(response):
+            gazette = self.parse_gazette_before_2018(response, pdf_link)
+            if self.start_date <= gazette["date"] <= self.end_date:
+                gazettes.append(gazette)
+
+        return gazettes
+
+    # FROM 2018 ONWARDS:
+
```
jvanz, line 98:
```suggestion
```

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,125 @@
...
+        for gazette in gazettes:
+            yield gazette
+
+    def months_interval(self):
+        month_rule = rrule(freq=MONTHLY, dtstart=self.start_date, until=self.end_date)
+        months = [month.date() for month in month_rule]
+        if months and months[-1].month != self.end_date.month:
+            months.append(self.end_date)
+        return months
```
jvanz, line 56: Sorry for my dumb question. But why do you need this `if`? 

```
>>> [month.date() for month in rrule(freq=MONTHLY, dtstart=datetime.date(2020, 1, 1), until=datetime.date.today())]
[datetime.date(2020, 1, 1), datetime.date(2020, 2, 1), datetime.date(2020, 3, 1), datetime.date(2020, 4, 1), datetime.date(2020, 5, 1), datetime.date(2020, 6, 1), datetime.date(2020, 7, 1), datetime.date(2020, 8, 1), datetime.date(2020, 9, 1), datetime.date(2020, 10, 1), datetime.date(2020, 11, 1)]
```

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,125 @@
...
+                    f"{self.URL_FROM_2018}{month.month}/{month.year}"
+                )
+
+    def parse(self, response):
+        gazettes = []
+
+        if self.URL_BEFORE_2018 in response.url:
+            gazettes = self.parse_month_before_2018(response)
+        else:
+            gazettes = self.parse_month_from_2018(response)
```
jvanz, line 46: You do not need this. To remove the entire `parse` method you can define the callback function when you return the request object in the `start_requests`. https://docs.scrapy.org/en/latest/topics/request-response.html?highlight=Request#scrapy.http.Request

`
callback (collections.abc.Callable) – the function that will be called with the response of this request (once it’s downloaded) as its first parameter. For more information see Passing additional data to callback functions below. If a Request doesn’t specify a callback, the spider’s parse() method will be used. Note that if exceptions are raised during processing, errback is called instead.
`

jvanz (changes requested):

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
fariaseduv: If not specified on CLI (-a parameter) the gazette will use the start_date = datetime.date(2011, 9, 30) [line 22]. The rrule will not include the current month if the today() is smaller than 30:

`[month.date() for month in rrule(freq=MONTHLY, dtstart=datetime.date(2011, 9, 30), until=datetime.date.today())][-3:]
`

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
fariaseduv: It's really better, thank you.

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
fariaseduv: So if someone define (with -a) start_date on day X of some month and today().day is smaller than X, the current month gazettes will be ignored.

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,109 @@
...
+                    callback = self.parse_month_before_2018
+                )
+            else:
+                yield scrapy.Request(
+                    f"{self.URL_FROM_2018}{month.month}/{month.year}",
+                    callback = self.parse_month_from_2018
+                )
+
+    def months_interval(self):
+        month_rule = rrule(freq=MONTHLY, dtstart=self.start_date, until=self.end_date)
```
rennerocha, line 40: **suggestion**: You can get rid of the unnecessary `if` statement if you change the value of `dtstart`:
```
month_rule = rrule(
    freq=MONTHLY, 
    dtstart=datetime.date(
        self.start_date.year, self.start_date.month, 1
    ), 
    until=end_date
)
```

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,109 @@
...
+        months = [month.date() for month in month_rule]
```
rennerocha, line 41: **suggestion (non-blocking)**: You could return a tuple with the values already formatted to be used by `start_requests`:

```
months = [(period.year, period.month) for period in month_rule]
```

An then you can simplify your code in `start_requests` with:
```
for year, month in self.months_interval():
   ...
```

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,109 @@
...
+    URL_BEFORE_2018 = "https://meriti.rj.gov.br/dom_mp/"
+    URL_FROM_2018 = (
+        "https://transparencia.meriti.rj.gov.br/diario_oficial_get.php?mesano="
+    )
+
+    start_date = datetime.date(2011, 9, 30)
+    end_date = datetime.date.today()
+
+    def start_requests(self):
+        janeiro_2018 = datetime.date(2018, 1, 1)
```
rennerocha, line 26: **nitpick**: Our code should be written preferably in English, so `january_2018` is a better name for this variable.

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,109 @@
...
+            months.append(self.end_date)
+        return months
+
+    def get_day_fom_pdf_link(self, pdf_link):
+        day_regex = r"[/-](\d{1,2})[^\d/][^/]*.pdf"
+        day = int(re.search(day_regex, pdf_link, re.IGNORECASE).group(1))
+        return day
+
+    def parse_gazette_before_2018(self, month_response, pdf_link):
+        year = int(month_response.url[-10:-6])
```
rennerocha, line 52: **suggestion**: You already have the year and the month when you create this request, so you could pass it as a `meta` argument or a `cb_kwargs`. This is more reliable as you don't rely on magic numbers to obtain this value, so if they change the URL (even one char is enough) this will break.

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,109 @@
...
+        date = datetime.date(year, month, day)
+
+        return Gazette(
+            date=date,
+            file_urls=[pdf_link],
+            is_extra_edition=False,
+            power="executive_legislative",
+        )
+
+    def get_pdf_links(self, month_response):
```
rennerocha, line 64: **question**: Why do you need to change the URL? We should keep the original URLs.

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+
+    def parse_month_from_2018(self, response):
+        gazettes = []
+        month_entries = json.loads(response.text)
+
+        for gazette_entry in month_entries:
+            gazette = self.parse_gazette_from_2018(gazette_entry)
+            if self.start_date <= gazette["date"] <= self.end_date:
+                gazettes.append(gazette)
```
rennerocha, line 107: **suggestion**: You don't need to store all gazettes in a list before returning them. Just yield them:
```
            if self.start_date <= gazette["date"] <= self.end_date:
                yield gazette
```

rennerocha (changes requested):
Added a few comments. Please verify them.

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
fariaseduv: That's right, this way we enforce that the end_date month will be included.

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
fariaseduv: Thank you, I'm trying to get used with the pythonic style.

data_collection/gazette/spiders/rj_sao_joao_de_meriti.py
fariaseduv: I didn't catch where the URL is being changed. Could you clear it for me?

fariaseduv (commented):
I miss click on "review" button so will get this opportunity to ask some questions: 

1. Do you plan to update CITIES.MD to reflect better the current census and vice-versa? https://censo.ok.org.br/get-data/
2. For other "associacao_municipios" spiders, you intend to expand it in each city file spider as was done with SC state?
3. Should I bring this question to some more approprieate place (discord, mail, issue etc) ?

rennerocha:
Closed as stale.

PR #347, opened 2020-11-03
Add base spider for doem.org.br and spider for Acajutiba/BA

rodbv:
Closes #300 

This is a new PR due to the several changes done on the project structure, check PR #330 for reference.

PR #377, opened 2020-11-13
Organize base spiders

rennerocha:
This idea was discussed during Python Brasil sprint. We have the base spiders spread in the structure of the project. Most of them were located in `base.py` , but we had other in a separated file `instar_base.py`. It is possible that we will have other cities that uses a common system that can be extracted as a base spider.

This PR wants to organize this to make it easier (and clearer) for a contributor where to place their new base spider (and related classes/functions).


giuliocc (approved):
Great job! Everything seems to be working :) LGTM

giuliocc:
#347 was merged and it has a base spider :/

Could you add that one too?

jvanz (approved):

jvanz:
> #347 was merged and it has a base spider :/
> 
> Could you add that one too?

Yeah, I think the branch needs to be rebased in the latest main and fix the missing base spider. ;)

jvanz (changes requested):
Marking as `Request changes` just to signal that we need to fix latest base spider.

PR #378, opened 2020-11-13
Add flake8 as a pre-commit step

rennerocha:
To make codebase cleaner, we should apply flake8 and solve the problems found there. Added it to pre-commit and fixed all existing problems in the project.


.pre-commit-config.yaml
```diff
@@ -16,3 +16,7 @@ repos:
     rev: 20.8b1
     hooks:
     -   id: black
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.4
+    hooks:
+    -   id: flake8
```
giuliocc, line 22: I feel like we could ignore `pycodestyle`'s errors since they can conflict with the other hooks. What do you think?

```suggestion
    -   id: flake8
        args: ['--ignore=E, W']
```

jvanz (approved):

.pre-commit-config.yaml
rennerocha: They are not conflicting with any other hook. I also think we should not ignore everything. I will just add a change to accept longer lines.

PR #379, opened 2020-11-14
Adopt Github Actions

anapaulagomes:
This PR adds `check` as a GitHub Action workflow and remove Travis (😈).

- [x] Replace Travis by GitHub Actions
- [x] Remove stale Docker workflow

We have considered running the spiders from a pull request ran by GitHub actions but we decided to not do it due to the [limitations of a free plan](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions) (some spiders take 10 hours running).


.github/workflows/tests.yml
```diff
@@ -0,0 +1,16 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Install dependencies
+        run: pip install -r data_collection/requirements.txt
```
giuliocc, line 14: Using cache could improve rerunning CI in the same PR. It should be tested if that's the case for us, though :)

```suggestion
      - name: Cache pip
        uses: actions/cache@v2
        with:
          path: ~/.cache/pip
          key: ${{ runner.os }}-pip-${{ hashFiles('data_collection/requirements.txt') }}
          restore-keys: |
            ${{ runner.os }}-pip-
      - name: Install dependencies
        run: pip install -r data_collection/requirements.txt
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,16 @@
+name: Run tests
+
+on: [ push, pull_request ]
```
giuliocc, line 3: What do you think about only running CI on PRs to `main`?

anapaulagomes: I'm fine with it but I think it limits us to only work with branches open against `main`. It may be a few cases where a person is working on something big (like multiple spiders) and want to split it in different PRs...

giuliocc (approved):
LGTM! :)

jvanz (approved):

PR #381, opened 2020-11-19
Add spider for São Paulo/SP

### LINKED ISSUE #380: Add spider for São Paulo - SP
According to CITIES.md there still isn't any spider for São Paulo, so I'll work in adding it.

The gazette can be found at http://www.docidadesp.imprensaoficial.com.br/Busca.aspx (as already tracked by the [census](https://censo.ok.org.br/andamento/?region=sudeste&page=4)).

#7 was regarding an alternative information source that isn't the gazette, and as mentioned there, it isn't the focus right now, so I'm creating this issue to track the progress on the spider to retrieve the gazette for São Paulo.

nfraprado:
Fixes #380.

Some caveats:

* I used the url scheme shown [here](https://github.com/okfn-brasil/querido-diario/issues/7#issuecomment-385122215) which directly gives the pdf, rather than serving a pdfViewer with no easy way to query the pdf file (at least I have no idea how to do it). Unfortunately, with this scheme the farthest back one can go is to July 1st, 2017.

* As already pointed in that same issue, this gazette serves each page as a separate PDF. I handled this by just getting the maximum page and returning the list of urls for each gazette (each one for a single page). But there a couple problems:
    * Since the pdf for each page is saved using a hash as its name, I think this might make it more difficult to later join the pages in the correct order. So it might be interesting to prepend the index number to the hash, like `1_<hash>.pdf`, etc. Or perhaps even simpler, have this information in the database if it isn't already.
    * Besides that, since each page requires a separate request, I need to do about 80 requests for each gazette, yielding a low requests ratio, hence the following error:
```
FAIL: Requests/Items Ratio/Ratio of requests over items scraped count
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/nfraprado/cur/querido_diario/querido-diario/data_collection/gazette/monitors.py", line 24, in test_requests_items_ratio
    self.assertLess(
AssertionError: 101.57142857142857 not less than 5 : 10157.14% is greater than the allowed 500%
                ratio of requests over items scraped.
```

This is my first time contributing to the project as well as using scrapy (besides the tutorial), so any feedback would be greatly appreciated :)

### DIFF
--- a/data_collection/gazette/spiders/sp_sao_paulo.py
+++ b/data_collection/gazette/spiders/sp_sao_paulo.py
@@ -1,8 +1,9 @@
 import locale
 import re
-from datetime import date, datetime, timedelta
+from datetime import date, datetime
 
 import scrapy
+from dateutil.rrule import DAILY, rrule
 
 from gazette.items import Gazette
 from gazette.spiders.base import BaseGazetteSpider
@@ -20,12 +21,9 @@ class SpSaoPauloSpider(BaseGazetteSpider):
     def start_requests(self):
         # Need to have the month's name in portuguese for the pdf url
         locale.setlocale(locale.LC_TIME, "pt_BR.UTF-8")
-        today = date.today()
-        day = self.start_date
-        while day < today:
+        for day in rrule(freq=DAILY, dtstart=self.start_date, until=date.today()):
             url = f"{self.BASE_URL}/nav_v6/header.asp?txtData={day.strftime('%d/%m/%Y')}&cad=1"
-            yield scrapy.Request(url, cb_kwargs=dict(day=day))
-            day = day + timedelta(days=1)
+            yield scrapy.Request(url, cb_kwargs=dict(day=day.date()))
 
     def get_max_page(self, response):
         page_txt = response.css("span.form-text::text").get()
--- a/data_collection/gazette/spiders/sp_sao_paulo.py
+++ b/data_collection/gazette/spiders/sp_sao_paulo.py
@@ -1,0 +1,54 @@
+import locale
+import re
+from datetime import date, datetime, timedelta
+
+import scrapy
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+RE_max_page_num = re.compile(r"\d+ de (\d+)")
+
+
+class SpSaoPauloSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3550308"
+    BASE_URL = "http://diariooficial.imprensaoficial.com.br"
+    allowed_domains = ["diariooficial.imprensaoficial.com.br"]
+    name = "sp_sao_paulo"
+    start_date = date(2017, 6, 1)
+
+    def start_requests(self):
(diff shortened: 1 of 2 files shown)


data_collection/gazette/spiders/sp_sao_paulo.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def start_requests(self):
+        # Need to have the month's name in portuguese for the pdf url
+        locale.setlocale(locale.LC_TIME, "pt_BR.UTF-8")
+        today = date.today()
+        day = self.start_date
+        while day < today:
+            url = f"{self.BASE_URL}/nav_v6/header.asp?txtData={day.strftime('%d/%m/%Y')}&cad=1"
+            yield scrapy.Request(url, cb_kwargs=dict(day=day))
+            day = day + timedelta(days=1)
```
jvanz, line 28: This is not wrong. But I think we can share other approaches as well. Other spiders use `rrule` to generate a date range. If you are interested, take a look. ;)

data_collection/gazette/spiders/sp_sao_paulo.py
```diff
@@ -0,0 +1,54 @@
+import locale
+import re
+from datetime import date, datetime, timedelta
+
+import scrapy
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+RE_max_page_num = re.compile(r"\d+ de (\d+)")
```
jvanz, line 10:
```suggestion
RE_MAX_PAGE_NUM = re.compile(r"\d+ de (\d+)")
```

jvanz (changes requested):

jvanz:
>       * Since the pdf for each page is saved using a hash as its name, I think this might make it more difficult to later join the pages in the correct order. So it might be interesting to prepend the index number to the hash, like `1_<hash>.pdf`, etc. Or perhaps even simpler, have this information in the database if it isn't already.

Good point. In this  SP case we can use the URL (stored in the database) to get the file order. However, I'm not sure if this will be extendable for other spiders in the future. We will probable will need a different structure to store the file order. @rennerocha  already suggested something like that in another spider.  But in that case, afaics, the order was not important. I do not believe this should prevent the PR to get merged. But it is something which should be in our backlog.  

>       * Besides that, since each page requires a separate request, I need to do about 80 requests for each gazette, yielding a low requests ratio, hence the following error:

Hmm, we will need a custom configuration when running this spider. 

> This is my first time contributing to the project as well as using scrapy (besides the tutorial), so any feedback would be greatly appreciated :)

Thanks for your contribution!

data_collection/gazette/spiders/sp_sao_paulo.py
nfraprado: Oh, indeed that's nicer. I'll make the change to use it shortly.

jvanz (approved):
Even if I do not believe this is ready for production because of the topic discussed in the PR, I think we can merge this. It will provide a starting point for further improvements. So, LGTM.

FYI: I've rebased the commits on the top of the main master.

PR #411, opened 2021-01-21
Add spider argument: end_date

gsusin:
I found it useful to have the end_date argument for testing purposes. So you can narrow the dates that your spider works on. 

Example of use:
`scrapy crawl sp_sao_bernardo_do_campo -a start_date=2016-06-01 -a end_date=2017-08-01`


jaswdr (approved):
LGTM

Issue #409, opened 2021-01-12
Add spider for São Bernardo do Campo - SP


gsusin:
According to CITIES.md there still isn't any spider for São Bernardo do Campo - SP, so I'll work in adding it.

The gazette can be found at https://www.saobernardo.sp.gov.br/web/sbc/todas-as-edicoes (as already tracked by the census). Their first data is from 2002-05-03 and they always use PDF.

trevineju:
Fechando a issue por já existir uma para o mesmo município

PR #412, opened 2021-01-21
Add spider for São Bernardo do Campo - SP

gsusin:
Issue #409


gsusin:
This spider downloaded 1404 files with 1175 different dates, from 2002-05-03 to 2021-01-15.

data_collection/gazette/spiders/sp_sao_bernardo_do_campo.py
```diff
@@ -0,0 +1,280 @@
...
+    end_date = date.today()
+
+    def start_requests(self):
+        base_url = "https://www.saobernardo.sp.gov.br/web/sbc/todas-as-edicoes"
+        yield scrapy.Request(base_url)
+
+    def parse(self, response):
+        base_files_url = "https://www.saobernardo.sp.gov.br"
+        years = response.xpath(
+            "/html/body/div[5]/div/div/div[1]/div/div[2]/div/div/section/div/div/div/a/text()"
```
anapaulagomes, line 24: You could reduce the XPath by using the classes that are above these fields in terms of hierarchy. Like:

```
div.portlet-content div.portlet-content-container div.portlet-body a
```
Apart from keeping the xpath short, it will also help with readability.

data_collection/gazette/spiders/sp_sao_bernardo_do_campo.py
```diff
@@ -0,0 +1,280 @@
...
+        ).re(r" *(.+)")
```
anapaulagomes, line 25: Since you're looking for years, would be better to have this regex as `\d{4}` or `\d+` (numbers only).

data_collection/gazette/spiders/sp_sao_bernardo_do_campo.py
```diff
@@ -0,0 +1,280 @@
...
+        hrefs = response.xpath(
+            "/html/body/div[5]/div/div/div[1]/div/div[2]/div/div/section/div/div/div/a/@href"
+        ).re(r"#(\w+)")
+        ids = dict(zip(years, hrefs))
+        duplicated = [" NM 1915"]
+
+        for y in range(self.start_date.year, self.end_date.year + 1):
```
anapaulagomes, line 32:
```suggestion
        for current_year in range(self.start_date.year, self.end_date.year + 1):
```

anapaulagomes (changes requested):
Hello @gsusin! Thank you for your contribution (and sorry about taking so long to review your code). 🚀 

I left a few comments but after taking a look at the extraction of the information I realized that the code could be simplified by reducing the amount of regex and repeated expressions. We value your time and work here but we have to make sure that the code is easy to understand and maintain by others too. What do you think of pairing on taking a look at it? Maybe I could help you to simplify some things and you could explain some decisions. If you also want to take the time to review some things again, it's fine as well. Just let me know.

anapaulagomes:
Ah, mais uma coisa: pode atualizar o [CITIES.md](https://github.com/okfn-brasil/querido-diario/blob/main/CITIES.md), por favor?

gsusin:
Thanks for your suggestions, @anapaulagomes! I updated the code to implement them. But if you think there are more improvements to do, I'm available to pair and clear any remaining points.

gsusin:
> Ah, mais uma coisa: pode atualizar o [CITIES.md](https://github.com/okfn-brasil/querido-diario/blob/main/CITIES.md), por favor?

Feito!

rennerocha:
Closed as stale.

PR #415, opened 2021-02-26
Refactor am_manaus to allow start_date

giuliocc:
Since there were two spiders inside one (executive and legislative)
a split was made. This simplifies things for the start_date
implementation but also for any other attribute that was specific
to only one of those spiders.

To make less requests in production (only last day's gazette), the
pagination was changed to step-by-step.

P.S. the am_manaus_legislative spider was not working and this
commit does not fix it.


jvanz:
Duplicate of #416

PR #417, opened 2021-02-27
Adding spider for Betim/MG

camilafracaro:
Created this spider with help from Giulio, Ana, Marcio and Yara. I didn't find an extra edition pattern and inside the gazette you could see that you have "ato do legislativo" and "ato do executivo" so we have both on the same gazette. I checked if it ran since 2008 and if it runs with only one page.


giuliocc (approved):
Hey, @camilafracaro ! Looks great to me! Sorry for the long delay :(

Added a commit to remove a comment that I thought was not worth asking from you since it was so minor and you already waited for so long :)

camilafracaro:
Thanks @giuliocc! What's the next step after approval?

giuliocc:
Nothing really, merging it now :)

Another great contribution! :heart:

PR #428, opened 2021-03-30
Fix selector to retrieve PDF URL for gazette

rennerocha:
The changed how the PDF URL is displayed.
After this PR is merged, we should run a full scrape for this city, so we can get gazettes that we missed in the last weeks.


giuliocc (approved):
Found no problems. LGTM :+1:

PR #437, opened 2021-04-25
Container local execution

andreformento:
@jvanz 
Possibility of run only with container

depends on https://github.com/okfn-brasil/querido-diario-data-processing/pull/19

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,131 @@
+data_collection/data
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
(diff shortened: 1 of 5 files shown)


jaswdr (approved):

rennerocha:
The scrapers can be executed using a simple Python virtualenv. In production we are using Scrapy Cloud, so I don't see any improvement in allowing to run inside a container. It is just another barrier for new developers with extra complexity.
If in future we decide to create our own containers to run the project in production, than this could be useful. Closing for now.

Issue #434, opened 2021-04-22
Add spider for Cianorte PR


ilzisilva:
Hello!
According to cities.md there is no spider for Cianorte-PR. The Official Gazette can be found here: 
https://cianorte.pr.gov.br/orgao-oficial
me and @Anafrozza are going to work on that.

PR #445, opened 2021-05-06
Spider cianorte

ilzisilva:
Issue #434


data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,31 @@
...
+    name = "pr_cianorte"
+    allowed_domains = ["cianorte.pr.gov.br"]
+    start_urls = ["https://cianorte.pr.gov.br/orgao-oficial/"]
+    current_page = 1
+    base_url = "https://cianorte.pr.gov.br/orgao-oficial/pagina/"
+
+    def parse(self, response):
+        lines = response.css("tr")[1:]
+        if len(lines):
+            for row in lines:
```
giuliocc, line 19: Just a common minor improvement to gain horizontal space:
```suggestion
        lines = response.css("tr")[1:]
        if not lines:
        	return
        	
        for row in lines:
        ...
```

data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,31 @@
...
+                date = row.xpath("td[2]/text()").extract()[0]
+                edition_number = row.xpath("td[1]/text()").extract()[0]
```
giuliocc, line 21: `.extract()[0]` can be reduced to `.get()`
```suggestion
                date = row.xpath("td[2]/text()").get()
                edition_number = row.xpath("td[1]/text()").get()
```

data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,31 @@
...
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class PrCianorteSpider(BaseGazetteSpider):
+    TERRITORY_ID = "4105508"
+    name = "pr_cianorte"
+    allowed_domains = ["cianorte.pr.gov.br"]
+    start_urls = ["https://cianorte.pr.gov.br/orgao-oficial/"]
+    current_page = 1
```
giuliocc, line 13: We could avoid using `current_page` as a class attribute by using [`cb_kwargs`](https://docs.scrapy.org/en/latest/topics/request-response.html#passing-additional-data-to-callback-functions).
```python
    def parse(self, response, current_page=1):
        ...
                yield Request(self.base_url + str(current_page), cb_kwargs=current_page+1)
```

data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,31 @@
...
+
+                yield Gazette(
+                    date=parse(date, languages=["pt"]).date(),
+                    edition_number=edition_number,
+                    file_urls=[row.css("a::attr(href)").get()],
+                    is_extra_edition=False,
+                    power="executive_legislative",
+                )
+                self.current_page += 1
+                yield Request(self.base_url + str(self.current_page))
```
giuliocc, line 31: Don't mind this too much, it's just personal preference but an f-string here would be nice and we wouldn't need a `base_url` class attribute
```suggestion
                yield Request(f"https://cianorte.pr.gov.br/orgao-oficial/pagina/{self.current_page}")
```

data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,31 @@
+from dateparser import parse
+from scrapy.http import Request
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class PrCianorteSpider(BaseGazetteSpider):
```
giuliocc, line 8: We should have a `start_date` attribute here. This will be used in production to capture gazettes in the last few days while not making unnecessary requests. In this spider, since we go back in time by paginating, you could verify if the gazette date you are extracting is before `self.start_date` and stop the crawl right there if you reach this point.

If in doubt, check some spiders for capitals, most of them already have `start_date`.

giuliocc (changes requested):
Hi @ilzisilva!

The spider is working great, from start to finish :) Awesome job!

Thank you very much for the PR! I've made some comments down below and also gave you some tips :)

Feel free to ask any questions. We're here to help and guide you through your first contribution :)

data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,34 @@
...
+    TERRITORY_ID = "4105508"
+    name = "pr_cianorte"
+    allowed_domains = ["cianorte.pr.gov.br"]
+    start_urls = ["https://cianorte.pr.gov.br/orgao-oficial/"]
+    current_page = 1
+    base_url = "https://cianorte.pr.gov.br/orgao-oficial/pagina/"
+
+    def parse(self, response):
+        lines = response.css("tr")[1:]
```
rennerocha, line 17: **suggestion**: You don't need to skip the first row, considering that the table has a well defined structure like:
```
<table>
  <thead>
    <tr><th>Table headers</th></tr>
  </thead>response.css("tr")[
  <tbody>
    <tr><td>Column you want</td></tr>
  </tbody>
</table>
```
You can use a more specific selector and get only the row inside `tbody` tag:
`lines = response.css("tbody tr")`

data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,34 @@
...
+        if not lines:
```
rennerocha, line 18: **suggestion**: You don't need to return if no lines are available. If the list is empty, the `for` loop next will just not be executed and the method will return.

data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,34 @@
...
+        	return
+        	
+        for row in lines:
+        ...
```
rennerocha, line 22: **nitpick**: These `...` are useless. You may remove it.

data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,34 @@
...
+                date = row.xpath("td[2]/text()").get()
+                edition_number = row.xpath("td[1]/text()").get()
+
+                yield Gazette(
+                    date=parse(date, languages=["pt"]).date(),
```
rennerocha, line 27: **suggestion (non-blocking)**: As the format of the dates is stable and consistent (DD/MM/YYYY), you could simply use Python standard library built-in [datetime.strptime](https://docs.python.org/3/library/datetime.html#datetime.datetime.strptime):

`date=datetime.strptime(date, "%d/%m/%Y")`

`dateparser` is an wonderful library, but it is not perfect and is more useful if you don't have a well-defined pattern for the dates you are handling.

data_collection/gazette/spiders/pr_cianorte.py
```diff
@@ -0,0 +1,34 @@
...
+                    edition_number=edition_number,
+                    file_urls=[row.css("a::attr(href)").get()],
```
rennerocha, line 29: **suggestion**: It is better to be more specific which link you want to collect:
`file_urls=row.xpath("td[3]/a/@href").getall()`
(I used `get_all` so the return is already a list)

data_collection/gazette/spiders/pr_cianorte.py
rennerocha: :+1: this is an issue that needs to be solved.

rennerocha (changes requested):
Looks good. Just added some suggestions for improvement. See @giuliocc comment about `start_date`. This is very important.

ilzisilva:
good night, thanks for all the tips, I made the changes that were requested.

giuliocc (changes requested):

data_collection/gazette/spiders/pr_cianorte.py
giuliocc: @rennerocha is totally right. I'm guessing you didn't find right because of the way the pagination is being made irregardless of having of not having `lines`.

Luckily, we can fix that with a [for-else](https://book.pythontips.com/en/latest/for_-_else.html#for-else). Just putting the request inside the `else` clause would make it work.

Also, a really minor thing that is totally just my preference in thinking it fits nicely: if we do it like this, we could change the `return` statement inside the loop to a `break`.

giuliocc (changes requested):

rennerocha:
Closed as stale.

PR #448, opened 2021-05-08
Add spider to Cariacica-ES

cecivieira:
The main decisions were:

- Make requests in a interval of 7 days;
- Scrape all item, even if it isn't an official gazette. For example: gazette cards appear with "oficial" in the title. So I was unsure if we should discard everything else and chose to scrape everything.

### DIFF
--- a/data_collection/gazette/spiders/es_cariacica.py
+++ b/data_collection/gazette/spiders/es_cariacica.py
@@ -1,0 +1,51 @@
+import datetime
+import re
+from urllib.parse import urlencode
+
+from dateutil.rrule import WEEKLY, rrule
+from scrapy.http import Request
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class EsCariacicaSpider(BaseGazetteSpider):
+    name = "es_cariacica"
+    TERRITORY_ID = "3201308"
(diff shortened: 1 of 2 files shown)


jaswdr:
LGTM

jaswdr (approved):

data_collection/gazette/spiders/es_cariacica.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class EsCariacicaSpider(BaseGazetteSpider):
+    name = "es_cariacica"
+    TERRITORY_ID = "3201308"
+    start_urls = ["https://www.cariacica.es.gov.br/publicacoes/diario-oficial/"]
+    start_date = datetime.date(2014, 7, 1)
+    end_date = datetime.date.today()
+
+    def parse(self, response):
```
rennerocha, line 19: **suggestion** Instead of raising the first requests in `parse`, you can use `start_requests` method. So you avoid a first useless request to the URL in `start_urls` list. [docs about this method here](https://docs.scrapy.org/en/latest/topics/spiders.html?highlight=start_requests#scrapy.spiders.Spider.start_requests)

rennerocha:
Closed as stale.

PR #403, opened 2020-12-13
Add Guaratinguetá-SP spider

tiagofer:
First draft for #360.

PR #451, opened 2021-05-17
Add Guaratinguetá-SP spider

jaswdr:
Fixes for PR #403

The original author seem to have abandoned the original PR, I've pick the changes, rebase it with the current main branch and fix the code, there were one invalid regex and a collection of "or" that I change to use the `any()`built in function.

### DIFF
--- a/data_collection/gazette/spiders/sp_guaratingueta.py
+++ b/data_collection/gazette/spiders/sp_guaratingueta.py
@@ -1,0 +1,54 @@
+import datetime
+import re
+
+import dateparser
+import scrapy
+from dateutil.rrule import YEARLY, rrule
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class SpGuaratinguetaSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3518404"
+    name = "sp_guaratingueta"
+    allowed_domains = ["https://guaratingueta.sp.gov.br/"]
+    start_date = datetime.date(2015, 6, 23)
+
+    def start_requests(self):
+        years = [
+            r.year
+            for r in rrule(
+                freq=YEARLY, dtstart=self.start_date, until=datetime.date.today()
+            )
+        ]
+        for year in years:
+            if year == datetime.date.today().year:
+                url = "https://guaratingueta.sp.gov.br/diario-oficial-da-estancia-turistica-de-guaratingueta/"
+            else:
+                url = f"https://guaratingueta.sp.gov.br/diario-oficial-{year}/"
+            yield scrapy.Request(url)
+
+    def parse(self, response):
+        texts = response.xpath(
+            '//*[@id="content"]/article/div[1]/ul/li/a//text()'
+        ).getall()
+        gazette_urls = response.xpath(
+            '//*[@id="content"]/article/div[1]/ul/li/a/@href'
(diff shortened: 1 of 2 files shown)


data_collection/gazette/spiders/sp_guaratingueta.py
```diff
@@ -0,0 +1,59 @@
...
+class SpGuaratinguetaSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3518404"
+    name = "sp_guaratingueta"
+    allowed_domains = ["https://guaratingueta.sp.gov.br/"]
+    start_date = datetime.date(2015, 6, 23)
+
+    def start_requests(self):
+        years = [
+            r.year
+            for r in rrule(
```
anapaulagomes, line 21: What's `r`? Can we give it a more expressive name? 🙏🏽

data_collection/gazette/spiders/sp_guaratingueta.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def parse(self, response):
+        texts = response.xpath(
+            '//*[@id="content"]/article/div[1]/ul/li/a//text()'
+        ).getall()
+        gazette_urls = response.xpath(
+            '//*[@id="content"]/article/div[1]/ul/li/a/@href'
+        ).getall()
+        for gazette_url, text in zip(gazette_urls, texts):
+            date = re.match("[0-9]{2}/[0-9]{2}/[0-9]{4}", text).group()
```
anapaulagomes, line 40: Some texts don't have a match for this pattern, causing the exception: `AttributeError: 'NoneType' object has no attribute 'group'`. Could you please look into this?

data_collection/gazette/spiders/sp_guaratingueta.py
```diff
@@ -0,0 +1,59 @@
...
+from dateutil.rrule import YEARLY, rrule
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class SpGuaratinguetaSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3518404"
+    name = "sp_guaratingueta"
+    allowed_domains = ["https://guaratingueta.sp.gov.br/"]
```
anapaulagomes, line 15: I think it is safe to remove this line, since `start_requests` is generating the ones you need. In case you want to keep it, you must adjust it to have only the domain. It will avoid this warning:

```
URLWarning: allowed_domains accepts only domains, not URLs. Ignoring URL entry https://guaratingueta.sp.gov.br/
```

anapaulagomes (changes requested):
Thanks for taking this one over, @jaswdr! 💯

jaswdr:
thanks @anapaulagomes for the review, will fix this today until the EOD

jaswdr:
@anapaulagomes fixed all comments, something that I notice is that there are duplicate entries on the website, and there are some few cases where it is used the same file in different dates, causing some constraint violations when trying to save them in the database. I didn't find a good way to handle this, since dropping the constraint is not a good way to fix it, and changing the checksum is also not good. What did you think can be a better approach here?

anapaulagomes (approved):
Hello @jaswdr! I've seen this problem in other spiders too and I don't know how to handle it. I'm going to open a discussion about this. Other than that, all good! 🚀

PR #443, opened 2021-04-27
Spider rj duque de caxias

HenriqueRicardoFigueira:
Author : Alexandre Yudi Kajihara

PR #102, opened 2018-08-03
Feature: Duque de Caxias/RJ Spider

danilolutz:
* Update CITIES.MD
* Update .gitignore for vscode users

Probably should be refactored or not...

Issue #66, opened 2018-05-31
Poor site of Duque de Caxias - RJ


AlvaroLQueiroz:
The site has two format of links and file name for the gazettes.
1. http://duquedecaxias.rj.gov.br/portal/boletim-oficial/2015/01-Janeiro/
2. http://duquedecaxias.rj.gov.br/portal/arquivos/2018/fevereiro/

The first format is for gazzetes under 2017 and is veeery incomplete. Is missing a LOT of gazzetes.

The second format is for gazzetes of 2017 and above. In this format there is no way to get the day of the gazzete.

It's necessary enter in contact with the city hall. What would be better, a project representative to contact the city hall or me?

cuducos:
> The second format is for gazzetes of 2017 and above. In this format there is no way to get the day of the gazzete.

The day is in the PDF, isn't it? If not, can you link a PDF missing the date?

Regarding the different URLs, que can have `start_urls` that yields the proper URL according to the date… does that make sense?

danilolutz:
Can I help with or assume this Spider?

AlvaroLQueiroz:
Yes @danilolutz , you can take this.

danilolutz:
Thanks a lot @AlvaroLQueiroz

rennerocha:
Closing stale issue.

PR #462, opened 2021-08-12
feat: rj duque de caxias spider

danilolutz:
Issue #66

@anapaulagomes i rewrote the whole spider in a new branch and i opened a new pull request. 

The previous one (#102 and maybe this one too: #443) can be closed!

Please, let me know if i need to refactor something or change some file i have forgotten!

### DIFF
--- a/data_collection/gazette/spiders/rj_duque_de_caxias.py
+++ b/data_collection/gazette/spiders/rj_duque_de_caxias.py
@@ -1,0 +1,128 @@
+import re
+from datetime import date, datetime
+
+import scrapy
+from dateparser import parse
+
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class RjDuqueDeCaxiasSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3301702"
+    name = "rj_duque_de_caxias"
+    allowed_domains = ["duquedecaxias.rj.gov.br"]
+    portal_url = "https://duquedecaxias.rj.gov.br/portal/{}.html"
+    directory_url = "https://duquedecaxias.rj.gov.br/portal/boletim-oficial/{}/{}"
+
+    months = [
+        "01-Janeiro",
+        "02-Fevereiro",
+        "03-Marco",
+        "04-Abril",
+        "05-Maio",
+        "06-Junho",
+        "07-Julho",
+        "08-Agosto",
+        "09-Setembro",
+        "10-Outubro",
+        "11-Novembro",
+        "12-Dezembro",
+    ]
+
+    def start_requests(self):
+        urls = [
+            {
+                "url": "https://duquedecaxias.rj.gov.br/portal/boletim-oficial.html",
+                "year": date.today().year,
+                "month": 1,
+            }
+        ]
+
+        for year in range(2013, 2017):
+            for month in self.months:
+                urls.append(
+                    {
+                        "url": self.directory_url.format(year, month),
+                        "year": year,
+                        "month": month[:2],
+                    }
+                )
+
+        for year in range(2017, date.today().year):
+            urls.append(
+                {
+                    "url": self.portal_url.format(year),
+                    "year": year,
+                    "month": None,
+                }
+            )
+
+        for url in urls:
+            yield scrapy.Request(
+                url["url"],
+                callback=self.parse,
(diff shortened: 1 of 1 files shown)


data_collection/gazette/spiders/rj_duque_de_caxias.py
```diff
@@ -0,0 +1,128 @@
...
+        if response.status == 200:
+            for element in response.xpath('//a[contains(@href,"pdf")]/@href'):
+                url = response.urljoin(element.get())
+
+                name = element.get().lower()
+
+                gazette_date = self.handle_date(response, element.get(), year, month)
+
+                extra_edition = self.is_extra_edition(year, name)
+
```
anapaulagomes, line 78: I believe that you don't need to skip the lines between these variables (line 71-79).

data_collection/gazette/spiders/rj_duque_de_caxias.py
```diff
@@ -0,0 +1,128 @@
...
+                    power="executive_legislative",
+                )
+
+    def is_extra_edition(self, year, name):
+        if year <= 2016:
+            extras = ["e", "a"]
+            return bool([ele for ele in extras if (ele in name)])
+        else:
+            extras = ["vol", "especial", "extra"]
+            return bool([ele for ele in extras if (ele in name)])
```
anapaulagomes, line 98: Sorry but what `ele` means?

```suggestion
            return bool([extra for extra in extras if extra in name])
```

data_collection/gazette/spiders/rj_duque_de_caxias.py
```diff
@@ -0,0 +1,128 @@
...
+        "06-Junho",
+        "07-Julho",
+        "08-Agosto",
+        "09-Setembro",
+        "10-Outubro",
+        "11-Novembro",
+        "12-Dezembro",
+    ]
+
+    def start_requests(self):
```
anapaulagomes, line 33: We need to support `start_date` and `end_date`. With the current approach is only possible to crawl all gazettes. Could you please add support for it?

data_collection/gazette/spiders/rj_duque_de_caxias.py
```diff
@@ -0,0 +1,128 @@
...
+    def is_extra_edition(self, year, name):
+        if year <= 2016:
+            extras = ["e", "a"]
+            return bool([ele for ele in extras if (ele in name)])
+        else:
+            extras = ["vol", "especial", "extra"]
+            return bool([ele for ele in extras if (ele in name)])
+
+    def handle_date(self, response, search_for, year, month):
+        gazette_date = None
```
anapaulagomes, line 101: Since you're creating the same variable inside the `if/else` you don't need this line here.

```suggestion
```

anapaulagomes (changes requested):
Thanks for your contribution! 🥇 I left a few comments but the most important one is about the date filter support. After you change this, I'll look the code again.

data_collection/gazette/spiders/rj_duque_de_caxias.py
danilolutz: Ok @anapaulagomes,  no problem! But for the years below 2017 I can't consider the days to filter, i have no way to get them, this is a problem?

anapaulagomes: Oh I see! What do you think of displaying a warning for this case and only filter by year? If the filtering is too much trouble, a warning could be just fine.

danilolutz: I can filter by months too. Only daily filtering is the problem.

How can I display this warning, are you have some example, please 😖?

anapaulagomes: Oh even better! :)

You can do it like this:

```
import warnings

warnings.warn("It isn't possible to filter by day. Using month and year.")
```
https://docs.python.org/3/library/warnings.html

danilolutz: Ok thanks @anapaulagomes

anapaulagomes:
Ah, mais uma coisa: pode atualizar o [CITIES.md](https://github.com/okfn-brasil/querido-diario/blob/main/CITIES.md), por favor?

rennerocha:
Closed as stale.

PR #470, opened 2021-08-27
Add spider for Americana

lcsvillela:
Full working spider for collect data from Americana


data_collection/gazette/spiders/sp_americana.py
```diff
@@ -0,0 +1,68 @@
...
+    name = "sp_americana"
+    allowed_domains = ["diariooficial.americana.sp.gov.br"]
+    start_date = datetime.date(2018, 9, 1)  # First gazette available
+    start_urls = ["https://diariooficial.americana.sp.gov.br/diario-oficial-edicaoAnterior.php"]
+    extra_editions = "https://diariooficial.americana.sp.gov.br/diario-oficial-edicaoExtra.php"
+
+    locations = {
+        "gazette": "//*[@class='day table-light']/strong/a",
+        "extra_gazette_url": "//*[@class='list-group-item']/a/@href",
+        "extra_gazette_date": "//*[@class='list-group-item']/text()"}
```
rennerocha, line 20: **issue** All code must be [correctly formatted](https://github.com/okfn-brasil/querido-diario/blob/main/CONTRIBUTING.md#automated-code-formatting) using Black. If you configure your [development environment](https://github.com/okfn-brasil/querido-diario#development-environment) with the `pre-commit` enabled, this will be applied automatically before you commit anything.

data_collection/gazette/spiders/sp_americana.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def parse(self, response):
+
+        param_url = f"?mes={self.start_date.month}&ano={self.start_date.year}"
+        base_url = response.url
+        url = base_url + param_url
+        date = self.start_date
+
+        while date.year < datetime.date.today().year or date.month < datetime.date.today().month:
```
rennerocha, line 29: **suggestion (non-blocking)** Instead of calculating `datetime.date.today()` every loop iteration, you could set a new auxiliary variable and calculate it only once:
```
today = datetime.date.today()
while <something>:
    <do_something>

File: data_collection/gazette/spiders/sp_americana.py [hunk 3]
```diff
@@ -0,0 +1,68 @@
...
+            yield response.follow(url, self.parse_gazette)
+            date = date + relativedelta(months=+1)
+            url = base_url + f"?mes={date.month}&ano={date.year}"
+
+        yield response.follow(self.extra_editions, self.extra_parse_gazette)
+
+    def parse_gazette(self, response):
+
+        gazettes = response.xpath(self.locations["gazette"]).extract()
```
rennerocha, line 38: **suggestion** [Scrapy docs](https://docs.scrapy.org/en/latest/topics/selectors.html#extract-and-extract-first) suggest that you use `get()/getall()` instead of `extract()/extract_first()`. All works, but we usually prefer to follow the same pattern.

data_collection/gazette/spiders/sp_americana.py
```diff
@@ -0,0 +1,68 @@
...
+class SpAmericanaSpider(Spider):
+    TERRITORY_ID = "3501608"
+    name = "sp_americana"
+    allowed_domains = ["diariooficial.americana.sp.gov.br"]
+    start_date = datetime.date(2018, 9, 1)  # First gazette available
+    start_urls = ["https://diariooficial.americana.sp.gov.br/diario-oficial-edicaoAnterior.php"]
+    extra_editions = "https://diariooficial.americana.sp.gov.br/diario-oficial-edicaoExtra.php"
+
+    locations = {
+        "gazette": "//*[@class='day table-light']/strong/a",
```
rennerocha, line 18: **suggestion** When searching elements by class, prefer the use of CSS selectors instead of XPath. Any change in the class definition may break this selector. For example, if we have `<td class="day table-light">` and for some reason it changes to `<td class="table-light day">`, your selector will break even if nothing else changes.

data_collection/gazette/spiders/sp_americana.py
```diff
@@ -0,0 +1,68 @@
...
+
+        while date.year < datetime.date.today().year or date.month < datetime.date.today().month:
+            yield response.follow(url, self.parse_gazette)
+            date = date + relativedelta(months=+1)
+            url = base_url + f"?mes={date.month}&ano={date.year}"
+
+        yield response.follow(self.extra_editions, self.extra_parse_gazette)
+
+    def parse_gazette(self, response):
+
```
rennerocha, line 37: **suggestion** Considering my previous comment about using CSS selector instead of XPath for class search, you could replace with this:

```
gazettes = response.css(".day a")  # This returns an iterator of selectors
for gazette in gazettes:
    file_url = gazette.css("::attr(href)").get()
```

data_collection/gazette/spiders/sp_americana.py
```diff
@@ -0,0 +1,68 @@
...
+        gazettes = response.xpath(self.locations["gazette"]).extract()
+        for gazette in gazettes:
+            file_url = gazette.split('"')[1]
+            date = gazette.split('"')[7][6:16]
```
rennerocha, line 41: **suggestion** Extracting data from strings using slices may be dangerous as a simple extra space in the string that can be manually added can break. For example, if they change `Data: 28/08/2021` with `Data:28/08/2021` (a very likely change if they add this information manually), your slice will not work properly. A regex is a bit more safe for this situation.

```
date = gazette.css("::attr(title)").re_first(r"(\d{2}\/\d{2}\/\d{4})")
```

This regex extracts from the string inside `title`, any string that has 2 digits followed by a `/`, then more two digits, another `/` and then 4 digits. So you will end with date equal to `03/08/2021` for example.

data_collection/gazette/spiders/sp_americana.py
```diff
@@ -0,0 +1,68 @@
...
+            date = dateparser.parse(date, date_formats=["%d/%m/%Y"]).date()
+            edition = re.findall(r'\d+', gazette.split(';')[2])[0]
```
rennerocha, line 43: **suggestion** Same with edition can be done without relying on right positions in a string:
```gazette.css("::attr(title)").re_first("No:\s*(\d+)")```

File: data_collection/gazette/spiders/sp_americana.py [hunk 8]
```diff
@@ -0,0 +1,68 @@
...
+                edition_number=edition,
+
+            )
+
+    def extra_parse_gazette(self, response):
+
+        gazettes = response.xpath(self.locations['extra_gazette_url']).extract()
+        dates = response.xpath(self.locations['extra_gazette_date']).extract()[2::3]
+        counter = 0
+        while counter < len(gazettes):
```
rennerocha, line 59: **suggestion** You don't need a `counter` nor handle with string slices:

```
extra_gazettes = response.css("li.list-group-item")
for extra_gazette in extra_gazettes:
   # each iteration will be handling one of the `<li>` that contains each extra gazette
```

data_collection/gazette/spiders/sp_americana.py
```diff
@@ -0,0 +1,68 @@
...
+            file_url = gazettes[counter]
+            date = dateparser.parse(dates[counter][2:12], date_formats=["%d/%m/%Y"]).date()
```
rennerocha, line 61: **issue** You are returning *all* extra gazettes, no matter the date. We can ignore the ones that are before `start_date`.
**issue** You are not collecting the edition number that is available.
**suggestion** to get the date and the edition number you could use different regexes.

rennerocha (changes requested):
@lcsvillela thanks for the PR. I also read your blog post explaining what you did and it was amazing! You did a great job! I added a few comments with suggestions and issues that I found in your PR. Please let me know if you didn't understand something.

data_collection/gazette/spiders/sp_americana.py
lcsvillela: Oh, I did not see about the Black, sorry. Too nice that! I was using Vim, only with pure PEP

data_collection/gazette/spiders/sp_americana.py
lcsvillela: really much better and clean, adjusted

data_collection/gazette/spiders/sp_americana.py
lcsvillela: The book I used to learn it's a little old, thanks for this information!

data_collection/gazette/spiders/sp_americana.py
lcsvillela: Using this construction it's like contains at XPath, too interesting!

data_collection/gazette/spiders/sp_americana.py
lcsvillela: adjusted :)

data_collection/gazette/spiders/sp_americana.py
lcsvillela: it's really better to use this characteristic!

data_collection/gazette/spiders/sp_americana.py
lcsvillela: I don't understand the first, why this is an issue? On the page have no gazette or extra gazette before start_date. In this case, would be better to adjust the start_date?

data_collection/gazette/spiders/sp_americana.py
lcsvillela: adjusted!

data_collection/gazette/spiders/sp_americana.py
lcsvillela: The issue about the date is resolved, adjusted.
The issue about edition number is resolved, adjusted :)

lcsvillela:
Ready for review again @rennerocha

PR #476, opened 2021-09-01
Update CITIES.md

diegobiavati:


rennerocha:
We need to update this automatically. https://github.com/okfn-brasil/querido-diario/issues/440 is discussing possible solutions.

PR #484, opened 2021-09-24
Belo Horizonte spider

rennerocha:


data_collection/gazette/spiders/mg_belo_horizonte.py
```diff
@@ -0,0 +1,54 @@
...
+
+    name = "mg_belo_horizonte"
+    allowed_domains = ["dom-web.pbh.gov.br"]
+    start_date = datetime.datetime(2000, 1, 2)
+    end_date = None
+
+    def start_requests(self):
+        base_url = "https://api-dom.pbh.gov.br/api/v1/edicoes/buscarpublicacaopordata"
+
+        end_date = self.end_date or datetime.date.today()
```
giuliocc, line 23: I'm guessing this was intended to not make a conflict with `scrapy crawl mg_belo_horizonte -a end_date=YYYY-MM-DD` right? If that's the only reason, I just checked and the base class is already overriding the class attribute if the argument is present
 
```suggestion
    end_date = datetime.date.today()

    def start_requests(self):
        base_url = "https://api-dom.pbh.gov.br/api/v1/edicoes/buscarpublicacaopordata"
```

data_collection/gazette/spiders/mg_belo_horizonte.py
```diff
@@ -0,0 +1,54 @@
...
+
+        rule_start_date = datetime.date(
+            self.start_date.year, self.start_date.month, self.start_date.day
+        )
+        date_list = list(rrule(freq=DAILY, dtstart=rule_start_date, until=end_date))
+        for date in date_list:
```
giuliocc, line 29: I'm not sure about this cast to `list`. Do we need it?

Also, the variable was created to improve legibility, right? You think we should avoid using `rrule` on its own?

giuliocc (changes requested):
<3

I've made some requests but I think it's pretty close to done :)

It is giving me code "429 Too many requests" on a lot of requests, could you test it with `DOWNLOAD_DELAY` increased or something like that?

Also, could you verify the style checks that didn't pass?

data_collection/gazette/spiders/mg_belo_horizonte.py
```diff
@@ -0,0 +1,54 @@
...
+from gazette.items import Gazette
+from gazette.spiders.base import BaseGazetteSpider
+
+
+class MgBeloHorizonteSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3106200"
+
+    name = "mg_belo_horizonte"
+    allowed_domains = ["dom-web.pbh.gov.br"]
+    start_date = datetime.datetime(2000, 1, 2)
```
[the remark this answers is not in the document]
rennerocha: > 1995-09-26

Thanks! I tried to find the oldest date, but you did a better job than me :-)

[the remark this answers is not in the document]
rennerocha: > Also, `start_date` as `datetime.datetime` is giving errors.
> 
> ```
>   File ".../querido-diario/raspadores/data_collection/gazette/pipelines.py", line 17, in process_item
>     if spider.start_date > item.get("date"):
> TypeError: can't compare datetime.datetime to datetime.date
> ```
> 
> This can be a `datetime.date` already, eliminating the need for a `rule_start_date` variable

`rrule` returns a `datetime`, so we need to convert it to `date` so it can be compared with `self.start_date` (as `date`). This was already fixed. I didn't understando why you get these errors, as they don't happen to me.

data_collection/gazette/spiders/mg_belo_horizonte.py
rennerocha: No reason. Maybe I was debugging the result and created the variable. I removed it as it won't be needed.

data_collection/gazette/spiders/mg_belo_horizonte.py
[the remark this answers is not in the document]
giuliocc: It wasn't happening when comparing `rrule` dates with `self.start_date` but in the pipeline when the `date` attribute in the `Gazette` item was compared against `self.start_date`. But now it's fixed, thanks :)

data_collection/gazette/spiders/mg_belo_horizonte.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class MgBeloHorizonteSpider(BaseGazetteSpider):
+    TERRITORY_ID = "3106200"
+
+    name = "mg_belo_horizonte"
+    allowed_domains = ["dom-web.pbh.gov.br"]
+    start_date = datetime.date(1995, 9, 26)
+    end_date = datetime.date.today()
+
```
giuliocc, line 18: When testing a full crawl I get status code 429 (Too many requests) as soon as 1998. And then most requests fail :(

I was able to solve it with some `DOWNLOAD_DELAY`, though...
```suggestion
    end_date = datetime.date.today()
    custom_settings = {"DOWNLOAD_DELAY": 0.5}
```

giuliocc (changes requested):
Everything else is fine :)

Issue #489, opened 2021-10-10
Add spider for Pau dos Ferros - RN


itepifanio:
The city wasn't added to the `CITIES.md` file yet. 

The official gazette can be found here: https://paudosferros.rn.gov.br/publicacoes.php

I'm working on the spider.

PR #494, opened 2021-10-16
Add Pau dos Ferros crawler

itepifanio:
Issue #489

## About This PR
It add a new crawler to the project. 

- Official gazette: https://paudosferros.rn.gov.br/publicacoes.php?cat=11&ta=3
- Territory id: 2409407

## Additional information
The current gazette has a problem that some of its items were duplicated (as you can see in the image below).

Screenshot from 2021-10-13 21-02-43

When the project pipeline tries to insert the gazette in the database it throws ` UNIQUE constraint failed: gazettes.territory_id, gazettes.date, gazettes.file_checksum ` since the database will not insert the same file again. As we discussed below it makes sense that this errors is throwing and handle it in the pipeline wouldn't be the best approach.


giuliocc:
@itepifanio avisa aí quando tiver pronto pra review :)

---

### English

@itepifanio tell me when it's ready for review :)

PR #523, opened 2021-11-02
Upgrade project libraries and dependencies

rennerocha:


anapaulagomes:
What about adopting dependabot?

PR #531, opened 2021-12-16
Enabling Brasilia-DF spider

rennerocha:
* Fix spider to use default `end_date` if not provided
* Enable it to be executed daily


jvanz (approved):
LGTM! :D

PR #535, opened 2022-01-13
Add spider for São Gonçalo / RJ

gsusin:
Based on work by @vitorbaptista and @chicocvenancio in PR#292, which is stalled for more than 1 year.


rennerocha:
Stale pull request.

PR #567, opened 2022-06-12
Update end date parameter to allow filter by period

rennerocha:
We need to allow to run the spider providing and `end_date` attribute, to avoid extra request for dates that we are not interested. By default `end_date` is set to the current date if not provided.


giuliocc (approved):
LGTM!

PR #568, opened 2022-06-12
Start using end_date parameter to filter by period

rennerocha:
To reduce the amount of data loaded, `end_date` use to decide the limits of the search needs to be configured.


giuliocc (approved):
### Português

Testado. Pegando todas as datas corretamente. LGTM!

---

### English

Tested. Every date is being scraped correctly. LGTM!

PR #589, opened 2022-07-04
Sobral-CE - Collect older gazettes

rennerocha:
We have available data from 10/06/1998 that weren't collected.


eerison:
in the website: https://queridodiario.ok.org.br/pesquisa?city=2312908

Sobral is with `nivel 1` is it correct? or is it a bug? as you added this spider for sobral, it should be `nivel 3`, isn't it?

trevineju:
yes. But it's also not returning anything from Sobral in the API. We'll check the infrastructure, thanks @eerison!

Issue #578, opened 2022-06-26
Verify connection timed out error in Mossoro-RN spider


rennerocha:
From time to time we are getting a timed out error when processing Mossoro-RN. Because of this, we are losing a few days of data.

[log_rn_mossoro_94.txt](https://github.com/okfn-brasil/querido-diario/files/8986680/log_rn_mossoro_94.txt)

eduardoplima:
I wasn't able to replicate the timeout error you reported, I actually got a 403 error when trying to scrape gazettes from Mossoro. It seems that the Mossoro gazette website has implemented a captcha verification page to block automated repeated requests, as can be seen on the attached file. I wonder if the solution for the timeout error solves both issues. 

![captcha_mossoro](https://user-images.githubusercontent.com/1484132/180625888-f8746d13-7fb8-4f8e-af2a-6530567c37db.png)

PR #645, opened 2022-09-26
Fix production errors in Mossoro-RN spider

rennerocha:


bmortella:
Is this related to #578 ?

